Issue #30, opened 2016-07-20
Remove KycoRawConnetionUp and KycoRawConnetionDown from event handler


beraldoleal:
On `core/event_handlers.py` we should remove any specific handle.

PR #38, opened 2016-07-22
Fix Issue #30 raw connections

diraol:


beraldoleal:
This PR is strange. It has a lot of commits from PR #37 . Should be only commits related issue to #30 .

diraol:
@beraldoleal Commits fixed! =)

kyco/core/events.py
```diff
@@ -12,37 +15,39 @@ class KycoEvent(object):
     Args:
         content (dict): Dictionary with all event informations
     """
-    def __init__(self, content):
+    def __init__(self, content, connection=None, timestamp=datetime.now()):
```
beraldoleal: This should be with timezone support. For instance:
`datetime.datetime.now(timezone.utc)`

kyco/core/switch.py
```diff
@@ -0,0 +1,134 @@
...
+from kyco.core.events import KycoSwitchUp
+
+log = logging.getLogger('Kyco')
+
+
+class KycoSwitch(object):
+    """This is the main class related to Switches modeled on Kyco."""
+
+    def __init__(self, switch_id, socket=None):
+        self.switch_id = switch_id
```
beraldoleal: switch_id = dpid, and remove old_ids.

diraol: @beraldoleal considering that we are going to use the dpid as the identifier of the switch, wouldn't be better if we use the dpid on the events instead of `connection` (`(ip, port)`) ?
I mean, we don't want to know the "connection" on the events, what we really need is a reference to the switch. right?

diraol:
@beraldoleal requested changes made.

PR #56, opened 2016-08-03
FIX #31 - KycoSwitch Class

### LINKED ISSUE #31: Create a Switch class on controller
We should create a Switch class to store connection and basic info about the switch.

diraol:
FIX #31


beraldoleal:
Please, try to use a commit description that makes sense with your changes.

PR #60, opened 2016-08-05
Issue #59 remove semaphore

### LINKED ISSUE #59: Remove unnecessary semaphore
@cemsbr has said that the `Queue` python module is both thread safe and have a blocking behavior when trying to get events from an empty queue.

So, the behavior that we implemented with the `Semaphore` (`Queue.Event`) is already built-in on the `Queue` class.

diraol:
Fix #59 - Removal of unnecessary Semaphore on Buffer class.
(Plus a commit with some improvements on exceptions, some logs and some pep8 fixes)

### DIFF
--- a/kyco/controller.py
+++ b/kyco/controller.py
@@ -369,8 +369,8 @@ class Controller(object):
         try:
             self.connections[connection_id]['socket'].send(message)
         except:
-            raise Exception('Error while sending a message to %s',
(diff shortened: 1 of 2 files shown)


beraldoleal:
Please, lets make the PR small as possible. Try to avoid use different subjects on the same PR.

Issue #58, opened 2016-08-04
Handle Client abrupt disconnection


diraol:
We need to define how to handle clients (switches) abrupt disconnections. This should be done with several checks (mainly on methods that communicate directly with the switch - send and read/recv).

Here we have two main cases on how to handle this.

The first one is when the disconnection happens before the end of the handshake process, when there is no `KycoSwitch` object "attached" to that connection (represented by the `connection_id`).

The second if when the disconnection happens after the end of the handshake process, when we have a `KycoSwitch` object "attached" to the connection.

How to create tests for this? Should we do them?

Issue #40, opened 2016-07-25
Log messages on exceptions


diraol:
Every time we have an exception it would be nice to log it also.

Considering this, we would log the exception at one of two places:
On the `init` of the called exception or on the except clause that got the exception.

beraldoleal:
Yes, I agree. I think that should be on the exception class.

hdiogenes:
Uncaught exceptions can be logged by defining a [handler function](https://stackoverflow.com/questions/8050775/using-pythons-logging-module-to-log-all-exceptions-and-errors) for [sys.excepthook](https://docs.python.org/3/library/sys.html?#sys.excepthook).

Issue #34, opened 2016-07-20
Create log handlers for core and napps


beraldoleal:
We should create a log handler for console and/or syslog, depending on how kyco is running.

For instance:

```
# daemon mode
$ ./kyco -d 
# All messages should be sent to syslog
```

```
# console mode
$ ./kyco
# All messages should be sent to console
```

This should include core logs and napps.

beraldoleal:
Also include a command on console to redirect logs to syslog.

beraldoleal:
This is related to #9

Issue #33, opened 2016-07-20
Rename loggers names to use context


beraldoleal:
`log = logging.getLogger('kytos[A]')` should be some thing like with context:

`log = logging.getLogger('kyco.napps.a')`

PR #66, opened 2016-08-22
Issue #58 abrupt disconnection

diraol:
@beraldoleal Now I think that this issue is solved. (at least this is what it looks like)

### DIFF
--- a/kyco/core/tcp_server.py
+++ b/kyco/core/tcp_server.py
@@ -67,7 +67,7 @@ class KycoOpenFlowRequestHandler(BaseRequestHandler):
         connection_id = (self.ip, self.port)
         event = KycoNewConnection(content=content, connection_id=connection_id)
         self.server.controller_put_raw_event(event)
-        log.debug("New connection from %s:%s", self.ip, self.port)
+        log.info("New connection from %s:%s", self.ip, self.port)
 
     def handle(self):
         curr_thread = current_thread()
@@ -77,9 +77,10 @@ class KycoOpenFlowRequestHandler(BaseRequestHandler):
             header = Header()
             binary_data = b''
 
-            raw_header = self.request.recv(8)
-            if not raw_header:
-                log.debug("Client %s:%s disconnected", self.ip, self.port)
+            try:
+                raw_header = self.request.recv(8)
+            except:
+                log.info("Client %s:%s disconnected", self.ip, self.port)
                 break
 
             while len(raw_header) < header_len:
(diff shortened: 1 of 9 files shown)


kyco/core/tcp_server.py
```diff
@@ -72,36 +72,41 @@ def setup(self):
...
+                    log.debug('Reading the binary_data')
+                    binary_data += self.request.recv(message_size)
+
+                # TODO: Do we need other informations from the network packet?
+                content = {'header': header, 'binary_data': binary_data}
+                connection_id = (self.ip, self.port)
+                event = KycoRawOpenFlowMessage(content=content,
+                                               connection_id=connection_id)
+                self.server.controller_put_raw_event(event)
+        except Exception:
```
beraldoleal, line 107: This exception is too generic, right ? If we have exceptions on line 106, 104 or 94 that means that the client is down ?

kyco/utils.py
```diff
@@ -124,6 +78,52 @@ def threaded_handler(*args):
...
+    controller_log.setLevel(logging.DEBUG)
+    controller_log.addHandler(controller_console_handler)
+
+    app_log = logging.getLogger('KycoNApp')
+    app_log.setLevel(logging.DEBUG)
+    app_log.addHandler(app_console_handler)
+
+    return controller_log
+
+
```
beraldoleal: I will accept this, but we should open an issue to configure the loggers from `logging.conf` file settings.

beraldoleal:
You don't need to fix this PR, but PLEASE, on the future, lets keep the PR only with commits that make sense to PR. For instance, the following commits should not be on this PR:

bec2227eea907cc55c52623d20fa61dabd011357
a0ec117ece796a52852987f1f4a3b905e5bb73e7
61218ab7c66b718cc9ed10f91aa00869f8daa82e
16e3068fe5fa4cb1882c317f613cb3f9234b2741 (part of)

I will wait to your comment/fix on the "too long exception" on `tcp_server.py` file, before accept.

Thanks for work on this.

kyco/utils.py
diraol: #23 
#33 
#34 
#40

kyco/core/tcp_server.py
diraol: Yes, this exception is quite generic (I think that a commit made after this one has changed the line 107 to a more proper exception).
Anyway, exceptions here would be only on lines 82 and 89.
I think that it would mean a client down, but you have said that it would be, for example, latency timeout.

PR #72, opened 2016-08-30
Improving setup.py and config to consider virtualenvs

diraol:


diraol:
This branch would still not pass on Travis.
This could be a problem related  to the `kytos/kyco-core-napps` dependency, that need to be updated and published into pypi with the fix that is waiting to be merged. (kytos/kyco-core-napps/pull/2)

PR #78, opened 2016-09-01
Improvements related to liveness napp

diraol:
Here we have some improvements on `Kyco` related to the `liveness` core napp.

Some changes were made to allow the development of the napp, and others are related to some fixes of bugs found after the implementation of the napp.

Some of the fixes are also related to the issue #58 also.

CHANGELOG:
- Improvements on client connection errors and exceptions handling
- Compatibilization with the liveness core napp


kyco/core/tcp_server.py
```diff
@@ -107,9 +109,18 @@ def handle(self):
...
                                                connection_id=connection_id)
                 self.server.controller.buffers.raw.put(event)
-        except Exception:
+        except (SocketError, OSError) as exception:
+            # TODO: Client disconnected is the only possible reason?
             log.info("Client %s:%s disconnected", self.ip, self.port)
-            raise Exception
+            error_content = {'destination': (self.ip, self.port),
+                             'exception': exception, 'event': None}
+            event = KycoMessageOutError(content=error_content)
```
beraldoleal: Rename KycoMessageOutError to KycoError or something like that

kyco/core/exceptions.py
```diff
@@ -3,6 +3,21 @@
...
+    def __str__(self):
+        return 'KycoCore exception: ' + super().__str__()
+
+
+class KycoSwitchOfflineException(Exception):
+    def __init__(self, dpid):
+        self.dpid = dpid
+
+    def __str__(self):
+        msg = 'The switch {} is offline, you can not send a message to it'
```
beraldoleal: Please, change this message to "The switch {} is not reachable. Please check the connection between the switch and the controller."

beraldoleal (changes requested):

kyco/core/tcp_server.py
diraol: Issue addressed on a191452

kyco/core/exceptions.py
diraol: Issue addressed on 6dc2f7e

PR #79, opened 2016-09-02
Fixes and small improvements

diraol:
This PR has some small changes related to:
- Source code organization
- Requirements
- PEP8 imports
- Use of now() accessory function.

It may be accepted after the PR #78, since it was developed considering the modifications from the branch used on that PR.

### DIFF
--- a/kyco/core/switch.py
+++ b/kyco/core/switch.py
@@ -87,9 +87,6 @@ class KycoSwitch(object):
     def is_connected(self):
         """Verifies if the switch is connected to a socket.
 
-        Try sending a null byte to the switch to check if the connection is
-        still alive.
-
         Returns:
             True: if the connection is alive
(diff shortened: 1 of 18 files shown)


kyco/core/switch.py
```diff
@@ -92,7 +92,9 @@ def is_connected(self):
             True: if the connection is alive
             False: if not.
         """
-        if (now() - self.lastseen).seconds > POOLING_TIME*2:
+        if self.socket is None:
+            return False
+        elif (now() - self.lastseen).seconds > CONNECTION_TIMEOUT:
             self.disconnect()
```
beraldoleal: `is_connected()` should only return true or false and not disconnect itself.

diraol: Fixed on e4883fa

PR #91, opened 2016-09-27
Improving napp unload

diraol:
Now the unload_napp method call the 'shutdown' method from the napp and
also remove the listeners from that method that are on the
events_listeners controller attribute.


diraol:
@beraldoleal Please, review this carefully. I am not fully sure if this improvement will work as expected, since the 'events_listeners' values contains the objects (or pointers) to the methods that will handle each event. I am not completely confidant that line 466 will work perfectly (but I think it will).

So, I just want another opinion on that.

Thanks.

beraldoleal:
It makes sense to me, but please test a napp unload, first.

diraol:
Napp unload working fine! =)

PR #99, opened 2016-10-05
Adding a message property on the KycoEvent class

diraol:


beraldoleal:
Even with travis failing I will accept because the error is not related.

PR #102, opened 2016-10-13
Fix signal kyco/core.shutdown

### LINKED ISSUE #101: Check controller stop action
Today controller.stop() should send a signall to all napps and shutdown his own thread. Maybe this is not working properly.

macartur:
- Fix signal shutdown in core.napps.py
- Add method _execute_loop to run execute method until the signal
  'kyco/core.shutdown' is received.
- Remove unused comments.
- Fix #101

### DIFF
--- a/kyco/core/napps.py
+++ b/kyco/core/napps.py
@@ -1,7 +1,7 @@
 import logging
 
 from abc import ABCMeta, abstractmethod
-from threading import Thread
+from threading import Event, Thread
 
 from kyco.utils import listen_to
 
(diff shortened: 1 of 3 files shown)


macartur:
After send a shutdown some times the thread can't stop. I will investigate it.

macartur:
@raphaelmcobe , I discovered the error, the package kyco-core-napps has a older implementation using signals with Classes and now we have a specific string to represent a signal. If you running locally all tests pass.

kyco/core/napps.py
```diff
@@ -48,12 +50,18 @@ def run(self):
...
+        self._execute_loop()
 
-    @listen_to('KycoShutdownEvent')
+    @listen_to('kyco/core.shutdown')
     def _shutdown_handler(self, event):
+        self.__event.set()
         self.shutdown()
 
+    def _execute_loop(self):
+        while(not self.__event.is_set()):
```
cemsbr: What if the user only want to run `execute()` once?

cemsbr (changes requested):
I'm taking care of the requested changes.

cemsbr (approved):

PR #113, opened 2016-11-04
Rename loggers #33

cemsbr:
We should use `getLogger(__name__)`. This closes the last milestone issue.

### DIFF
--- a/kyco/core/buffers.py
+++ b/kyco/core/buffers.py
@@ -6,7 +6,7 @@ from kyco.core.events import KycoEvent
 
 __all__ = ['KycoBuffers']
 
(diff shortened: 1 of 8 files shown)


kyco/core/buffers.py
```diff
@@ -6,7 +6,7 @@
 
 __all__ = ['KycoBuffers']
 
-log = logging.getLogger('Kyco')
+log = logging.getLogger(__name__)
```
cemsbr, line 9: Using `__name__` is considered a good convention in [Python docs](https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial).

Issue #106, opened 2016-10-28
Fix CI tests


cemsbr:
Tests have been failing for several days. It seems to be a Pypi package update issue.

PR #114, opened 2016-11-04
Fix ci #106

cemsbr:


cemsbr:
Fix and speed up CI builds. After this PR and kytos/kyco-core-napps#38, new builds should work.

Basically, it installs python-openflow and kyco-core-napps from the develop branch, so it does not depend on Pypi packages anymore. As the test breaks with certain combinations of apps, only of.core app is installed, enough for the tests.

PR #122, opened 2016-11-08
[core] Add interface speed in JSON

cemsbr:


cemsbr:
Tests are broken since https://github.com/kytos/kyco/commit/7e8db53f158df490b9d94995618b5c5356286da6

PR #124, opened 2016-11-08
Fix get_speed when interface.feature is None

macartur:


cemsbr:
Thanks, but this was fixed in #125.

PR #135, opened 2016-11-15
Fix tests

macartur:


macartur:
@diraol, I made changes in api_server from controller to use existent  _thread.

PR #146, opened 2016-12-03
Review class and methods from kyco

macartur:


cemsbr:
@beraldo, please, this is important to #137.

PR #150, opened 2016-12-08
Kytos sphinx template

diraol:


cemsbr:
Nice work. Only two issues in the attached screenshot (main page, scroll down everything, bottom part removed):
- In the top-left corner, there's something I can't see;
- It's difficult to read our partners.

screenshot

diraol:
@cemsbr that's weird, on both points (top left and bottom right) there should be the logos from kytos project and from our partners.

PR #210, opened 2017-01-10
Napp management refactor

diraol:
- Adds the NApps directory into the system path, so we can access the napps as python libraries;
 - Improve the "Load Napp" method;
 - Adds "Enable" and "Disable" napps methods.

### DIFF
--- a/kyco/config.py
+++ b/kyco/config.py
@@ -6,6 +6,7 @@ will be overridden by the option on command line.
 """
 
 import os
+import sys
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from configparser import ConfigParser
 
@@ -16,6 +17,8 @@ if 'VIRTUAL_ENV' in os.environ:
 else:
     BASE_ENV = '/'
 
+# include napps modules
+sys.path.append(os.path.join(BASE_ENV, 'var/lib/kytos'))
 
 class KycoConfig():
     """KycoConfig class handle settings of Kyco."""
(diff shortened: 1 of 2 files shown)


kyco/controller.py
```diff
@@ -564,3 +564,22 @@ def unload_napps(self):
         for napp_name in list(self.napps):
             if not isinstance(self.napps[napp_name], KycoCoreNApp):
                 self.unload_napp(napp_name)
+
+    def enable_napp(self, napp_author, napp_name):
+        """Enable a NApp by creating the needed symbolic link on the system."""
+
+        napp_abs_path = os.path.join(self.options.napps, '.installed',
```
beraldoleal: this vars should be more global.

kyco/controller.py
```diff
@@ -565,6 +565,18 @@ def unload_napps(self):
...
+
+        napp_sym_path = os.path.join(self.options.napps,
+                                     napp_author, napp_name)
+
+        if os.path.exists(napp_sym_path):
+            os.remove(napp_sym_path)
+            log.info('The NApp %s/%s disabled.', napp_author, napp_name)
+        else:
+            log.warning('NApp %s/%s was not enabled.', napp_author, napp_name)
+
```
beraldoleal, line 589: May be is more pythonic with try... except

kyco/config.py
```diff
@@ -16,6 +17,8 @@
 else:
     BASE_ENV = '/'
 
+# include napps modules
+sys.path.append(os.path.join(BASE_ENV, 'var/lib/kytos'))
```
beraldoleal: `os.path.join(BASE_ENV, 'var/lib/kytos')` is already been used on many parts of code. Please make this a var or config.

beraldoleal (changes requested):

cemsbr:
Note: build failed :(

PR #213, opened 2017-01-20
Fix build, bug, add coveralls

cemsbr:
Tests are fixed and improved, as in python-openflow. An admin needs to enable coveralls.


cemsbr:
@beraldoleal, tests are failing without this PR.

PR #219, opened 2017-02-04
Nappmgr

cemsbr:
- Smaller controller: some methods moved to NAppsManager and kytos-utils
- Improve websocket shutdown
- Improve tests

### DIFF
--- a/kyco/controller.py
+++ b/kyco/controller.py
@@ -532,7 +532,7 @@ class Controller(object):
 
     def load_napps(self):
         """Load all NApps enabled on the NApps dir."""
-        napps = NAppsManager(self.options)
+        napps = NAppsManager(self.options.napps)
         for author, napp_name in napps.get_enabled():
             try:
                 log.info("Loading NApp %s/%s", author, napp_name)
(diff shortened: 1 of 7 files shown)


kyco/controller.py
```diff
@@ -501,106 +502,71 @@ def add_new_switch(self, switch):
...
-                self.events_listeners.pop(event_type)
+        napp = self.napps.pop((author, napp_name), None)
+        if napp is None:
+            log.warn('NApp %s/%s was not loaded', author, napp_name)
+        else:
+            napp.shutdown()
+            # Removing listeners from that napp
+            for event_type, napp_listeners in napp._listeners.items():
+                event_listeners = self.events_listeners[event_type]
+                for listener in listeners:
```
diraol: @cemsbr I think that there is something wrong with the '*listeners*' variables on this block...

kyco/napps_manager.py
```diff
@@ -0,0 +1,25 @@
+"""Manage Network Application files."""
+from os import listdir, path
+
+
+class NAppsManager:
+    """Deal with NApps at filesystem level."""
+
+    def __init__(self, options):
```
diraol: Why not passing the 'folder' itself instead of the full options? [This would allow us to instantiate multiple Managers and/or avoid the need of instantiation of the KycoConfig class.

diraol (changes requested):

kyco/controller.py
cemsbr: Yes, of course! It should be `napp_listeners`.

kyco/napps_manager.py
cemsbr: Nice, I'll do it.

PR #234, opened 2017-02-21
Removing references to "KycoCoreNApps"

### LINKED ISSUE #232: Remove references to "KycoCoreNApp"
All NApps now have the same "status", there are no more "CoreNApps".

diraol:
FIX #232

### DIFF
--- a/kyco/core/napps.py
+++ b/kyco/core/napps.py
@@ -7,7 +7,7 @@ from kyco.utils import listen_to
 
 log = logging.getLogger(__name__)
 
-__all__ = ('KycoCoreNApp', 'KycoNApp')
+__all__ = ('KycoNApp')
 
 
 class KycoNApp(Thread, metaclass=ABCMeta):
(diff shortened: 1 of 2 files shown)


kyco/core/napps.py
```diff
@@ -7,7 +7,7 @@
 
 log = logging.getLogger(__name__)
 
-__all__ = ('KycoCoreNApp', 'KycoNApp')
+__all__ = ('KycoNApp')
```
cemsbr, line 10: A valid tuple here would be `__all__ = ('KycoNApp',)` or simply `__all__ = 'KycoNApp',` (always with comma).

cemsbr (changes requested):
We also depend on removing KycoCoreNApp references in kyco-core-napps for the tests to pass.

kyco/core/napps.py
diraol: @cemsbr, the KycoCoreNapps parte is on another PR. :)

cemsbr:
Dependency: kytos/kyco-core-napps#170

kyco/core/napps.py
cemsbr: I noticed later and made a reference.

PR #235, opened 2017-02-21
Avoid multiple calls of NApp shutdown method

### LINKED ISSUE #233: Multiple NApp shutdown calls
When shutting down the controller, the NApps shutdown method is being called multiple times (3).

This is due to the fact that we dispatch a KycoShutdownEvent for each buffer, and then the NApps are being notified for this Shutdown event from each buffer.

The expected behavior is to be called only once.

diraol:
FIX #233

### DIFF
--- a/kyco/core/napps.py
+++ b/kyco/core/napps.py
@@ -86,8 +86,9 @@ class KycoNApp(Thread, metaclass=ABCMeta):
         Paramters
             event (:class:`KycoEvent`): event to be listened.
         """
-        self.__event.set()
-        self.shutdown()
+        if not self.__event.is_set():
+            self.__event.set()
+            self.shutdown()
 
     @abstractmethod
     def setup(self):


kyco/core/napps.py
```diff
@@ -86,8 +87,10 @@ def _shutdown_handler(self, event):
         Paramters
             event (:class:`KycoEvent`): event to be listened.
         """
-        self.__event.set()
-        self.shutdown()
+        if not self.__shutdown_already_called:
+            self.__shutdown_already_called = True
```
cemsbr: Can we avoid creating a new variable by using the event itself like below?
```python
if not self.__event.is_set():
    self.__event.set()
    self.shutdown()
```

cemsbr (changes requested):

kyco/core/napps.py
beraldoleal: @diraol  ?

diraol: Done as requested. I liked  the suggestion @cemsbr

PR #242, opened 2017-03-11
Improving Kytos-core

macartur:
Changing websockets to socketio.
Adding APIServer class.


kyco/controller.py
```diff
@@ -92,49 +91,13 @@ def __init__(self, options):
         self.log_websocket = LogWebSocket()
         self.console_websocket = ConsoleWebSocket(controller=self)
 
-        self.app = Flask(__name__)
+        self.app = APIServer(__name__)
```
diraol: I am not sure if we should use `app` as the attribute name for the "APIServer". It does not seems to make sense to me. It may be a more meaningful name, like "api_server", "api", "rest_server", "rest" or something like this. Any other naming suggestions?

macartur: OK, I guess we can to use api_server.

beraldoleal: @macartur waiting...

PR #243, opened 2017-03-12
Fixing travis build to use new repositories

macartur:


beraldoleal:
Thanks but @cemsbr already did.

PR #287, opened 2017-03-27
Refactoring KytosNapps log register.

macartur:


diraol (approved):
For me thats a much better solution than the one we had before..

PR #289, opened 2017-03-28
Fixing from kytos.core.core import Controller

macartur:


bin/kytosd
```diff
@@ -5,7 +5,7 @@ import signal
 import daemon
 
 from kytos.core.config import KytosConfig
-from kytos.core.core import Controller
+from kytos.core import Controller
```
beraldoleal, line 8: `kytos.core.controller import Controller` right ?

beraldoleal (changes requested):
Also we have to change on all documents because this is a critical change.

bin/kytosd
beraldoleal: Ok, is right this way. Thanks.

beraldoleal (approved):

Issue #293, opened 2017-03-30
Setup error caused by imports


diraol:
Now that we are importing some classes on our `kytos/core/__init__.py` file (such as `from kytos.core.controller import Controller`), and the `__init__.py` is read on the `setup.py`, we are facing a "recursive import" error such as the one described here https://milkr.io/kfei/5-common-patterns-to-version-your-Python-package. It is not a real recursion error, but while loading `__init__.py` the `controller.py` is also loaded, and it imports Flask, that has not been installed yet [it will be installed during the setup].

PR #294, opened 2017-03-30
#293 Fix broken setup

diraol:


kytos/core/config.py
```diff
@@ -9,7 +9,7 @@
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from configparser import ConfigParser
 
-from kytos.core import __version__
+from kytos.core.metadata import __version__
```
beraldoleal:
```
from kytos.core import metadata
```
And when using:

`metadata.__version__`

setup.py
```diff
@@ -126,19 +125,22 @@ def create_path(self, file_name):
 
 requirements = [i.strip() for i in open("requirements.txt").readlines()]
 
+meta_file = open("kytos/core/metadata.py").read()
+metadata = dict(re.findall("__([a-z]+)__\s*=\s*'([^']+)'", meta_file))
```
beraldoleal: With my previous comment, we do not need this.

kytos/core/config.py
beraldoleal: Besides that I think that we should think about this more carefully. Looks like `__version__` needs to be a module thing. PEP0440. That is good, but please, let's think on a better solution.

beraldoleal (changes requested):

kytos/core/config.py
diraol: I've implemented with _metadata, to avoid _metadata being always exposed to the kytos users (it is not necessary).

diraol: I've looked at PEP440 and some others and I haven't found any imposition of file to hold this information. But, since I agree that it would be nice to have it on the module, I've imported the version from _metadata into the `__init__` file.

setup.py
diraol: Done

beraldoleal (approved):

PR #297, opened 2017-03-30
Adding LogManager class to handle kytos's logs

### LINKED ISSUE #34: Create log handlers for core and napps

macartur:
Fix #34

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -25,7 +25,8 @@ from flask import Flask, request
 from kytos.core.api_server import APIServer
 from kytos.core.buffers import KytosBuffers
 from kytos.core.events import KytosEvent
-from kytos.core.helpers import now, start_logger
+from kytos.core.helpers import now
+from kytos.core.log import LogManager
 from kytos.core.napps_manager import NAppsManager
 from kytos.core.switch import Switch
 from kytos.core.tcp_server import KytosOpenFlowRequestHandler, KytosServer
@@ -89,7 +90,8 @@ class Controller(object):
 
         self.websockets = {}
 
-        self.log = start_logger(__name__)
+        self.log = None
+        self.log_manager = None
 
         self.api_server = APIServer(__name__)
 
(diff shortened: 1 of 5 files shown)


kytos/core/controller.py
```diff
@@ -25,7 +25,8 @@
 from kytos.core.api_server import APIServer
 from kytos.core.buffers import KytosBuffers
 from kytos.core.events import KytosEvent
-from kytos.core.helpers import now, start_logger
+from kytos.core.helpers import now
+from kytos.core.log_manager import LogManager
```
beraldoleal: My only concern is here. I think that is better if it was something, like:

```
from kytos.core.log import LogManager
```
Or

```
from kytos.core.log.manager import LogManager
```

beraldoleal (changes requested):

kytos/core/log_manager.py
```diff
@@ -0,0 +1,85 @@
...
+        """Contructor of LogManager.
+
+        This method will create a LogManager instance with the mode given.
+
+        Args:
+            mode(string): Type of log that will be registered by kytos.
+        """
+        logging.basicConfig(format=self.fmt, level=getattr(logging, mode))
+
+    def new_logger(self, logger_name=None):
```
diraol: What if no `logger_name` is passed? It may be better to set this attribute (`logger_name`) as mandatory.

kytos/core/log_manager.py
```diff
@@ -0,0 +1,85 @@
...
+            logger_name(string): logger name that will be instantiated.
+        Returns:
+            logger(`logging.Logger`): instance of logging.Logger.
+        """
+        if logger_name not in self.loggers.keys():
+            logger = logging.getLogger(logger_name)
+            self.loggers[logger.name] = logger
+            self.__add_handlers_enabled_to_logger(logger)
+
+        return logger
```
diraol: There is an `else` clause missing here. If `new_logger()` is called with a `logger_name` that already exists this method cannot return `logger` since it does not exist. Then, I think this method could be:
```
if logger_name not in self.loggers.keys():
    logger = logging.getLogger(logger_name)
    self.loggers[logger.name] = logger
    self.__add_handlers_enabled_to_logger(logger)
    return logger
else:
    return self.loggers.get(logger_name)
```

kytos/core/controller.py
diraol: Thinking about our discussions to reduce imports sizes, we could do:
```
from kytos.core import LogManager
```
can't we?

diraol (changes requested):

kytos/core/controller.py
beraldoleal: @diraol, I think that put everything in core is a bad idea. Controller ok. But Log lets split into log module.

kytos/core/log_manager.py
beraldoleal: @diraol good point.

cemsbr (changes requested):
Please, use pylama path/to/your/files. There are linter errors that are ignored by current python setup.py test.

cemsbr:
Status: me and @diraol are fixing these issues.

cemsbr:
Closed in favor of #314

Issue #300, opened 2017-03-31
Move existing and update blueprints from wiki


erickvermot:
Move existing blueprints from wiki to kytos/docs/blueprints, and rename them sequentially.

diraol:
@erickvermot and @beraldoleal can we close this issue?

erickvermot:
I think we could update their references, structure and markup before closing...

beraldoleal:
All blueprints were moved ? If yes, we can close.

erickvermot:
All I could find on the wikis of each project.
I still have to delete the wikis though. I have not deleted yet because I wasn't sure if you would want to keep track of the past changes on the blueprints.

beraldoleal:
@erickvermot no, you can delete the wikis also. No problem.

erickvermot:
wikis deleted. closing the issue.

PR #301, opened 2017-03-31
moving blueprints from wiki

erickvermot:
refers to #300


erickvermot:
after the PR, the blueprints should be reviewed and updated.
I'll open a new issue for each one and ref this issue for tracking.

beraldoleal (approved):

Issue #328, opened 2017-04-05
Bug with setuptools


diraol:
Depending on your system setup and how you are installing python3.6 and/or creating the virtualenv, sometimes some dependencies from `setuptools` are not correctly installed. It has something to do with the version of setuptools installed and also with pip.

There are three packages that usually present problems regarding this: `appdirs`, `packagin` and `six`.

The easiest solution that I have found seems to be pin setuptools on the requirements. This will upgrade setuptools on the environment and, then, all dependencies will be installed.

Some references about this issue:
 - https://github.com/pypa/setuptools/pull/933
 - http://setuptools.readthedocs.io/en/latest/history.html#v34-0-0
 - https://github.com/ActiveState/appdirs/issues/89
 - https://www.reddit.com/r/Python/comments/5pwngp/setuptools_34_has_been_released_and_breaks_with/
 - https://github.com/pypa/setuptools/issues/951

cemsbr:
I'm working on #317 and it may help.

beraldoleal:
I fixed this on last commit, but since that I was not aware of this issue it was not closed automatically. Thanks.

PR #329, opened 2017-04-05
kytos/kytos#328 - Fix bug with setuptools

diraol:


beraldoleal:
Thanks @diraol, but I just fixed this too.

PR #335, opened 2017-04-05
Adding configuration templates using jinja2

macartur:


.gitignore
```diff
@@ -101,3 +101,6 @@ aux_*.sh
 node_modules
 include
 kytos/web-ui/source/static/css/*
+
+# ignore templates
+*.conf
```
diraol: I wouldn't ignore *.conf. May be just `etc/kytos/kytos.conf`.

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
         call('make -C docs clean', shell=True)
 
 
-class DevelopMode(develop):
+class CommonInstall:
+    """Class with common method used by children classes."""
+
+    @staticmethod
+    def generate_file_from_template(templates,
+                                    destiny=os.path.dirname(__file__),
```
diraol: s/destiny/destination/

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+                                    **kwargs):
+        """Write the template file using the appropriated path.
```
diraol: > Create a config file based on a template file.

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+
+        If no destiny is passed the file will be create next to the template
```
diraol: > If no destination is passed, the new conf file will be created on the directory of the template file.

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+        file.
+
+        Args:
+            template (string): Path of the template origin.
```
diraol: > Path of the template file

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+            destiny (string): Directory where the template will be created.
```
diraol: s/destiny/destination

> Directory in which the config file will be placed.

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+        """
+        for path in templates:
+            with open(path, 'r', encoding='utf-8') as src_file:
+                content = Template(src_file.read()).render(**kwargs)
+                dst_path = os.path.join(destiny, path.replace('.template', ''))
```
diraol: s/destiny/destination/

setup.py
```diff
@@ -90,7 +90,44 @@ def run(self):
...
+                with open(dst_path, 'w') as dst_file:
+                    dst_file.write(content)
+
+
+class InstallMode(install, CommonInstall):
+    """Class used to overwrite the default installation using setuptools.
+
+    This class will write the templates used by Kytos SDN Controller.
```
diraol: > Besides doing the default install, the config files used by Kytos SDN Platform will also be created, based on templates.

setup.py
```diff
@@ -100,24 +137,34 @@ class DevelopMode(develop):
...
+    def create_paths():
+        """Method used to create the paths used by Kytos in develop mode."""
+        directories = [os.path.join(BASE_ENV, 'etc/kytos')]
+        for directory in directories:
+            if not os.path.exists(directory):
+                os.makedirs(directory)
+
 
 requirements = [i.strip() for i in open("requirements.txt").readlines()]
 napps_dir = os.path.join(BASE_ENV, 'var/lib/kytos/napps/.installed')
```
diraol: This information should be provided by the conf file, doesn't it?

I mean, I am not sure if we should handle `napps_dir` during `kytos` setup, since it may be changed based on the kytos.conf file. So, both in kytos and kytos-utils we should check the `napps_dir` when needed and create it if it does not exist, right? @beraldoleal does it make sense? Should this point be addressed on another issu?

diraol (changes requested):

.gitignore
beraldoleal: `kytos.conf` should not exist anymore, right ?

diraol: You are right, but when installing in "development mode" (`python setup.py develop`) the installer should create the `kytos.conf` file, put it on the `etc/kytos` directory (on the repository dir) and do the symbolic link to this directory. So, we will have a `kytos.conf` file under the git tree. Therefore adding it to gitignore seems to be important to avoid commiting it by mistake. =)

setup.py
beraldoleal: The problem is that at this point, the user has not configured it yet (during the setup).

So I think that is correct. We will replace the {{ prefix }} with the same BASE_ENV. So it will make consistent. Makes sense ?

diraol: I understand your point.

But we are creating this directory at this point because something would break later on while using the project.

Now consider that we have created this directory during the setup, and then the user changes it on the generated `kytos.conf` file (for example to `/tmp/napps_dir/`). Then, something on kytos or kytos-utils will be broken, since the folder does not exists.

So, what I am saying is that every "code" that reads from the `kytos.conf` and may use this directory should check if it exist prior to using it.

(By the way, this has made me think about the relationship between `kytos.conf` [kytos project] and `~/.kytosrc`, [kytos-utils project]. Shouldn't they be aligned somehow?)

beraldoleal: I see your point. If the user changes the config, for sure is his/her responsibility to change the napps. We should not create napps dirs here.

~/.kytosrc is regarding the client. Only the client.

/etc/kytos/kytos.conf is about the daemon.

On the future kytosrc (client) will have only connection to controller config. And all other settings is returned from REST API.

diraol: Ok, we have a consensus after a cup of coffee =)

`kytos` (`kytosd`) will never be responsible for creating NApps related directories. This will be done by `kytos` (`kytos-utils`) [`NappsManager` class]. So, we can remove this line (170) from here.

If kytos (`kytosd`) does not find the `napps_dir` defined in `kytos.conf`, then it must assume that it is the same case as if there is no NApp installed. On those cases, `kytosd` must log a Warning message about it.

diraol (approved):

cemsbr (changes requested):
@macartur , we have to import jinja2 after installing "install_requires". Move the import to first line of `CommonInstall.generate_file_from_template`. If the linter complains about it, add `  # noqa` in the end of the line. Please, also add a comment above it so anybody can understand why this import is not in the expected section.

cemsbr (approved):

PR #350, opened 2017-04-17
Sets the appropriate permission to the napps folder.

### LINKED ISSUE #349: Napps installed module directory permission bug
similar to:
https://github.com/kytos/kytos-utils/issues/77
Installed Napps module directory is being created with wrong permissions, and makes the napp invisible to regular users.

erickvermot:
closes #349

### DIFF
--- a/kytos/core/napps/manager.py
+++ b/kytos/core/napps/manager.py
@@ -1,6 +1,7 @@
 """Manage Network Application files."""
 import json
 import logging
+import os
 import shutil
 import urllib
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


diraol:
@erickvermot Shouldn't we do this on all directories under the "napps_dir" ? (`kytos`, `kytos/napps`, `kytos/napps/.installed`, `kytos/napps/<author>`, `kytos/napps/<author>/<napp>`?

erickvermot:
@diraol, as I remember the _create_module function creates the parent directories with the correct permissions. The only folder that presents the problem is the one created using mkdtemp and later copied preserving the original 700 mode set by mkdtemp: The ``<napp>`` folder.

erickvermot:
* The ``<napp>`` folder.

diraol (approved):
Got it! =)

PR #353, opened 2017-04-17
Adding Controller.restart

### LINKED ISSUE #103: Create controller restart action
Today the Kyco controller don't have a restart action.

macartur:
Fix #103

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -1,6 +1,6 @@
 """Module used to handle a API Server."""
+import logging
 import os
-
 from urllib.error import URLError
 from urllib.request import urlopen
 
@@ -11,7 +11,7 @@ from flask_socketio import SocketIO
 class APIServer:
     """Api server used to provide Kytos Controller routes."""
 
-    def __init__(self, app_name):
+    def __init__(self, app_name, debug=False):
         """Contructor of APIServer.
 
         This method will instantiate a server with SocketIO+Flask.
@@ -21,9 +21,22 @@ class APIServer:
         """
         dirname = os.path.dirname(os.path.abspath(__file__))
         self.flask_dir = os.path.join(dirname, '../web-ui/source')
+        self.log = logging.getLogger('werkzeug')
+        self.set_debug(debug)
 
         self.app = Flask(app_name, root_path=self.flask_dir)
-        self.server = SocketIO(self.app)
+        self.server = SocketIO(self.app, async_mode='threading')
+
+    def set_debug(self, debug=False):
+        """Method used to set debug mode.
+
+        Args:
+            debug(bool): Boolean value to turn on/off debug mode.
+        """
+        if debug is True:
+            self.log.setLevel(logging.DEBUG)
+        else:
+            self.log.setLevel(logging.WARNING)
 
     def run(self, *args, **kwargs):
         """Method used to run the APIServer."""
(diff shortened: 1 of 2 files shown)


kytos/core/api_server.py
```diff
@@ -23,7 +27,7 @@ def __init__(self, app_name):
         self.flask_dir = os.path.join(dirname, '../web-ui/source')
 
         self.app = Flask(app_name, root_path=self.flask_dir)
-        self.server = SocketIO(self.app)
+        self.server = SocketIO(self.app, async_mode='threading')
```
diraol: @macartur why have you set this `async_mode='threading'`? What do this attribute changes?

kytos/core/api_server.py
```diff
@@ -1,12 +1,16 @@
...
+import logging
 import os
-
 from urllib.error import URLError
 from urllib.request import urlopen
 
 from flask import Flask, request, send_from_directory
 from flask_socketio import SocketIO
 
+# OMIT API server logs
```
diraol: 1) Why omitting the API server logs?
2) Shouldn't this be a config from our controller (to be set on `kytos.conf`)?

kytos/core/controller.py
```diff
@@ -169,6 +170,18 @@ def register_rest_endpoint(self, *options, **kwargs):
         """Method used to return the endpoints registered by APIServer."""
         self.api_server.register_rest_endpoint(*options, **kwargs)
 
+    def restart(self, graceful=True):
+        """Restart Kytos SDN Platform.
```
diraol: Actually this will restart a Kytos SDN Controller, not 'Platform'.

kytos/core/controller.py
```diff
@@ -169,6 +170,18 @@ def register_rest_endpoint(self, *options, **kwargs):
...
+
+        Args:
+            graceful(bool): Represents the way that Kytos will restart.
+        """
+        if self.started_at is not None:
+            self.stop(graceful)
```
diraol, line 180: The `stop()` may take a while (since there are lot's of threads to be stopped). I think that we may need to use a `callback` behavior here, waiting the stop to finish it's procedures and just then starting the controller again.

@beraldoleal any ideas on this?

diraol (changes requested):

kytos/core/controller.py
beraldoleal: @diraol we can accept like this and on the future improve the shutdown process.

kytos/core/api_server.py
macartur: OK, API server will get the debug variable from `kytos.conf`.

kytos/core/api_server.py
macartur: How the API server is running in thread, we need to use `async_mode` in SocketIO server, to start and shutdown the server.

kytos/core/controller.py
macartur: Fixed.

kytos/core/controller.py
cemsbr: I didn't understand how we can make the stop process faster only by using callbacks. I agree to wait when this becomes a problem.

cemsbr (approved):

kytos/core/controller.py
macartur: The method `stop_controller` will wait all threads finish before finish the method.

beraldoleal:
@diraol your changes request are ok ?

diraol:
Yes they are @beraldoleal!

diraol (approved):

kytos/core/controller.py
diraol: The proposal was not to make the stop process faster, but to wait for it to finish before proceeding to a new start. But, as @macartur said, the stop method will wait until everything was stopped prior to return. Thus, the behavior that I "requested" is already there. =)

PR #355, opened 2017-04-18
fixes napp folder permission by avoiding using mkdtemp.

erickvermot:
Also includes random number in tmp folder name.

cleaner solution to:
https://github.com/kytos/kytos/issues/349

### DIFF
--- a/kytos/core/napps/manager.py
+++ b/kytos/core/napps/manager.py
@@ -1,7 +1,6 @@
 """Manage Network Application files."""
 import json
 import logging
-import os
 import shutil
 import urllib
 from pathlib import Path
@@ -90,7 +89,6 @@ class NAppsManager:
             dst = self._installed / napp.username / napp.name
             self._create_module(dst.parent)
             shutil.move(str(napp_folder), str(dst))
-            os.chmod(dst, mode=0o755)
         finally:
             if pkg_folder and pkg_folder.exists():
                 shutil.rmtree(str(pkg_folder))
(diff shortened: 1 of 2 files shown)


kytos/core/napps/manager.py
```diff
@@ -90,7 +89,6 @@ def install(self, napp_uri, enable=True):
             dst = self._installed / napp.username / napp.name
             self._create_module(dst.parent)
             shutil.move(str(napp_folder), str(dst))
-            os.chmod(dst, mode=0o755)
```
beraldoleal, line 93: If I understood, we are:

1. Creating a tmp dir without fixing a proper permission (each user has your own umask);
2. Moving this folder to destination;

So yes, we need to force a chmod here, because we cannot make sure what is the tmp permissions, right ?

beraldoleal (changes requested):

kytos/core/napps/manager.py
erickvermot: The tmp permission is set considering the user umask. I know this does not force a desired permission, but this is the way all the other folders are being created, including the parents, like the module's directory created by _create_module(). The only folder that was not created in this way was the one created with mkdtemp().
I was just trying to keep the folder consistent with the others.
You are right, but if some mode is desired to be forced (755 for ex) we should force chmod on all the other methods too.

erickvermot: Should I include a chmod line in the creation of the other directories too?

beraldoleal: @erickvermot For now, I think that do not set this will cause a problem in some cases because of strange users' umask.

But we can leave with that for now, and wait until this problem happen again. I will accept this as it is now.

PR #358, opened 2017-04-19
Updating documentation

diraol:


diraol:
This PR is breaking because of an update on `pycodestyle`, that is catching "new" erros on our codebase. The test failure has nothing to do with the changes made by this PR.

PR #360, opened 2017-04-19
Create lockfile to assert single instance execution of kytosd

### LINKED ISSUE #340: Abort when another instance is running
Today, if one instance is already started we should abort.

There is a try exception but is not working since that we are working with threads.

renanrodrigo:
Fix #340

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -119,6 +119,7 @@ class Controller(object):
 
     def start(self):
         """Create lockfile and call start_controller method."""
+        self.enable_logs()
         pid = os.getpid()
 
         # Checks if a lockfile exists. Creates a new file.
@@ -134,8 +135,9 @@ class Controller(object):
                 os.kill(old_pid, 0)
                 # If kill() doesn't returns an error, it is still running.
                 # Otherwise, overwrite the file and proceed.
-                self.log.info("Failed to create a lockfile."
-                              "Is kytos running?")
+                self.log.info("Failed to create a lockfile. "
+                              "Is kytos already running?")
+                self.log.info("Aborting")
                 exit(1)
             except OSError:
                 lockfile = open(LOCK_PATH, mode='w')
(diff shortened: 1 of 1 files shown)


renanrodrigo:
Depends on #361

kytos/core/controller.py
```diff
@@ -33,6 +33,8 @@
 
 __all__ = ('Controller',)
 
+LOCK_PATH = '/tmp/lockytos'
```
beraldoleal: This should be configured on `/etc/kytos/kytos.conf`. And should have a default value to `/var/run/kytos.pid`

kytos/core/controller.py
```diff
@@ -116,6 +118,34 @@ def enable_logs(self):
         self.log = logging.getLogger(__name__)
 
     def start(self):
+        """Create lockfile and call start_controller method."""
+        pid = os.getpid()
+
+        # Checks if a lockfile exists. Creates a new file.
+        try:
+            lockfile = open(LOCK_PATH, mode='x')
```
beraldoleal: s/lockfile/pidfile/g

kytos/core/controller.py
```diff
@@ -116,6 +118,34 @@ def enable_logs(self):
...
+            # running.
+            try:
+                existing_file = open(LOCK_PATH, mode='r')
+                old_pid = int(existing_file.read())
+                os.kill(old_pid, 0)
+                # If kill() doesn't returns an error, it is still running.
+                # Otherwise, overwrite the file and proceed.
+                self.log.info("Failed to create a lockfile."
+                              "Is kytos running?")
+                exit(1)
```
beraldoleal: Replace 1 with a proper constant. For instance `os.EX_CANTCREAT`

kytos/core/controller.py
```diff
@@ -212,6 +243,7 @@ def stop_controller(self, graceful=True):
         self.started_at = None
         self.unload_napps()
         self.buffers = KytosBuffers()
+        os.remove(LOCK_PATH)
```
beraldoleal: s/LOCK_PATH/PID_FILE/g.

Also you have to handle exceptions like `FileNotFoundError`

beraldoleal (changes requested):

renanrodrigo:
Rebased and working

PR #369, opened 2017-04-20
Create folder for pidfile on setup

### LINKED ISSUE #368: Enhance pidfile creation to support execution as non-root user
Since we are generating a pidfile in `/var/run`, `kytosd` needs root permission to be executed outside a virtual environment. Need to check pidfile creation to support common users running kytosd.

renanrodrigo:
Fix #368.

Warning: To solve this problem, a folder is being created with 777 permission inside /var/run. If this is not desired somehow, please request changes.

### DIFF
--- a/etc/kytos/kytos.conf.template
+++ b/etc/kytos/kytos.conf.template
@@ -9,7 +9,7 @@ workdir = {{ prefix }}/var/lib/kytos
 
 # PID file to write. When the controller starts, it will save the his pid on
 # this file.
-pidfile = {{ prefix }}/var/run/kytosd.pid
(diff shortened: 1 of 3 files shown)


beraldoleal:
Let's do on this way, for 2017.1b2. And then on the future we create a proper user/group during install.

diraol (approved):

cemsbr:
A more secure mode is like /tmp's: anyone can create files inside it, but only the owner can delete them (or root). /tmp's octal is 1777.

diraol:
@renanrodrigo conflicts....

PR #372, opened 2017-04-24
Fix #370 - jinja2 on setup

### LINKED ISSUE #370: Load Jinja2 during kytos setup
After solving kytos/kytos#357 we still have an error during our setup.

`Jinja2` is a requirement of our project (it is in requirements.txt), and is correctly installed by the `setup` procedure (`python setup.yp install`).
But as it is also used/loaded on the setup script, we are facing a `timing` problem.

The available (installed) modules from `PYTHONPATH` are loaded on the beginning of the `setup.py` script. During this script, all requirements are installed, including the `Jinja2` package. But the available modules (`sys.path`) are not reloaded, thus `Jinja2` is not available to be used inside the `setup.py` script, which raises an error during the install procedure.

If the setup is run again, everything works fine (since `Jinja2` was installed on the first run).

So, we need  to figure out how to load (import) `jinja2` on the `setup.py` script "on the runtime".

diraol:


setup.py
```diff
@@ -97,6 +98,26 @@ class CommonInstall:
...
+        """Add Jinja2 into sys.path.
+
+        As Jinja2 may be installed during this setup, it will not be available
+        on runtime to be used here. So, we need to look for it and append it on
+        the sys.path.
+        """
+        #: First we find the 'site_pkg' directory
+        site_pkg = None
+        for path in sys.path:
+            if PurePosixPath(path).stem == 'site-packages':
```
cemsbr: We can use `Path` here to have a shorter import and a more portable code.

setup.py
```diff
@@ -97,6 +98,26 @@ class CommonInstall:
...
+                site_pkg = Path(path)
+                break
+
+        #: Then we get the 'Jinja2' egg directory and append it into the
+        #: current sys.path.
+        jinja2path = site_pkg.glob('Jinja2*').__next__()
```
cemsbr: `next(site_pkg.glob('Jinja2*'))`

cemsbr (changes requested):

setup.py
diraol: I thought (because of the documentation) that Path didn't had the stem property. I've checked and it have, so I'm changing it.

setup.py
```diff
@@ -110,7 +131,9 @@ def generate_file_from_template(templates,
             destination (string): Directory in which the config file will
                                   be placed.
         """
-        from jinja2 import Template
+        CommonInstall.add_jinja2_to_path()
+        from jinja2 import Template  # pylint: disable=import-error
```
beraldoleal:
```
try:
  from jinja2 import Template
except ImportErrorBlabal:
  ...add_jinja2_to_path()
```

Issue #292, opened 2017-03-30
Fix tcp_server


erickvermot:
Currently the tcp_server is strictly OpenFlow, and it does some of the parsing inside it's ``HandlerClass.handle`` method. This needs to change so that the controller can operate with different protocols, and protocol versions.
From what I understand about the wanted structure, OF parsing should be done by ``of_core`` napp, not by ``kytos/core`` or inside the handler. If this is the way to go, some things need to change:

- implement the tcp server so that it is initialized with a generic ``KytosRequestHandler`` instead of a specific Openflow one.
- implement the ``RequestHandler`` so that it triggers a ``kytos/server.data.in`` event with the raw bytes read from the socket.
- make the ``Connection`` hold info about it's state (new, setup, established, finished), and it's protocol.
- process the raw bytes received via the ``server.data.in`` inside ``of_core``. It includes slicing the raw data flow into packets, and parsing the initial hello message before the protocol version is set by the negotiation.

I believe the TCP server and the handler could be moved outside the core, to a new napp or inside ``of_core``, as was also suggested by @diraol. 

Maybe the Handler could contain ``protocol_pack`` and ``protocol_unpack`` functions set once negotiation is completed, that should be used to parse the incoming messages.

diraol:
@erickvermot can we close this issue?

erickvermot:
IDK, should we close it now, or should we close it when we move the whole TCP Server and handler to of_core? Or should we open another issue for that?
@beraldoleal @diraol

beraldoleal:
This is a big issue, right ? So far I understood, we fixed part of this with @erickvermot PR. So I suggest to open a new issue, just to move tcp_server to outside core.

beraldoleal:
I'm closing this issue, if there is any issue here that is not splited as new issue, please open a new specific issue.

PR #374, opened 2017-04-25
New request handler

erickvermot:
refers to #292.
depends on PR https://github.com/kytos/kytos-napps/pull/217

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -29,8 +29,8 @@ from kytos.core.events import KytosEvent
 from kytos.core.helpers import now
 from kytos.core.logs import LogManager
 from kytos.core.napps.manager import NAppsManager
-from kytos.core.switch import Switch
-from kytos.core.tcp_server import KytosOpenFlowRequestHandler, KytosServer
+from kytos.core.switch import CONNECTION_STATE, Switch
+from kytos.core.tcp_server import KytosRequestHandler, KytosServer
 from kytos.core.websocket import LogWebSocket
 
 __all__ = ('Controller',)
@@ -178,7 +178,7 @@ class Controller(object):
         self.log.info("Starting Kytos - Kytos Controller")
         self.server = KytosServer((self.options.listen,
                                    int(self.options.port)),
-                                  KytosOpenFlowRequestHandler,
+                                  KytosRequestHandler,
                                   self)
 
         raw_event_handler = self.raw_event_handler
@@ -366,10 +366,22 @@ class Controller(object):
 
             message = triggered_event.content['message']
             destination = triggered_event.destination
-            if destination:
-                destination.send(message.pack())
+            if (destination and
+                    not destination.state == CONNECTION_STATE.FINISHED):
+                packet = message.pack()
+                destination.send(packet)
+                self.log.debug('Connection %s: OUT OFP, ' +
+                               'version: %s, type: %s, xid: %s',
+                               destination.id,
+                               message.header.version,
+                               message.header.message_type,
+                               message.header.xid)
+                self.log.debug(packet.hex())
                 self.notify_listeners(triggered_event)
                 self.log.debug("MsgOutEvent handler called")
+            else:
+                self.log.info(
+                    "connection closed. Cannot send message")
 
     def app_event_handler(self):
         """Handle app events.
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -25,11 +25,12 @@ from threading import Thread
 from kytos.core.api_server import APIServer
 from kytos.core.buffers import KytosBuffers
 from kytos.core.config import KytosConfig
+from kytos.core.connection import CONNECTION_STATE
 from kytos.core.events import KytosEvent
 from kytos.core.helpers import now
 from kytos.core.logs import LogManager
 from kytos.core.napps.manager import NAppsManager
-from kytos.core.switch import CONNECTION_STATE, Switch
+from kytos.core.switch import Switch
 from kytos.core.tcp_server import KytosRequestHandler, KytosServer
 from kytos.core.websocket import LogWebSocket
 
--- a/kytos/core/switch.py
+++ b/kytos/core/switch.py
@@ -1,17 +1,13 @@
 """Module with main classes related to Switches."""
 import json
 import logging
-from enum import Enum
-from errno import EBADFD, ENOTCONN
-from socket import error as SocketError
-from socket import SHUT_RDWR
 
 from pyof.v0x01.common.phy_port import PortFeatures
 
 from kytos.core.constants import CONNECTION_TIMEOUT, FLOOD_TIMEOUT
 from kytos.core.helpers import now
 
-__all__ = ('Interface', 'Connection', 'Switch')
+__all__ = ('Interface', 'Switch')
 
 log = logging.getLogger(__name__)
 
@@ -192,110 +188,6 @@ class Interface(object):
             json (string): Json filled with interface attributes.
         """
         return json.dumps(self.as_dict())
-
-
-class CONNECTION_STATE(Enum):
-    """Enum of possible general connections states."""
-
-    NEW = 0
-    SETUP = 1
-    ESTABLISHED = 2
-    FAILED = 3
-    FINISHED = 4
-
-
-class ProtocolInfo:
-    """Class to hold simple protocol information for the connection."""
-
-    def __init__(self):
-        self.name = None
-        self.version = None
-        self.state = None
-
-
-class Connection(object):
-    """Connection class to abstract a network connections."""
-
-    def __init__(self, address, port, socket, switch=None):
-        """The constructor method have the below parameters.
-
-        Parameters:
-          address (HWAddress): Source address.
-          port (int): Port number.
-          socket (socket): socket.
-          switch (:class:`~.core.switch.Switch`): switch with this connection.
-        """
-        self.address = address
-        self.port = port
-        self.socket = socket
-        self.switch = switch
-        self.state = CONNECTION_STATE.NEW
-        self.protocol = ProtocolInfo()
-        self.remaining_data = b''
-
-    @property
-    def state(self):
-        """Return the state of the connection."""
-        return self._state
-
-    @state.setter
-    def state(self, new_state):
-        if new_state not in CONNECTION_STATE:
-            raise Exception('Unknown State', new_state)
-        self._state = new_state # noqa
-        log.info('Connection %s changed state: %s',
-                 self.id, self.state)
-
-    @property
-    def id(self):
-        """Return id from Connection instance.
-
-        Returns:
-            id (string): Connection id.
-        """
-        return (self.address, self.port)
-
-    def send(self, buffer):
-        """Send a buffer message using the socket from the connection instance.
-
-        Parameters:
-            buffer (bytes): Message buffer that will be sent.
-        """
-        try:
-            if self.is_connected():
-                self.socket.sendall(buffer)
-        except (OSError, SocketError) as exception:
-            log.debug('Could not send packet. Exception: %s', exception)
-            self.close()
-
-    def close(self):
-        """Close the socket from connection instance."""
-        if self.socket:
-            log.debug('Shutting down Connection %s', self.id)
-            try:
-                self.socket.shutdown(SHUT_RDWR)
-                self.socket.close()
-            except OSError as e:
-                if e.errno not in (ENOTCONN, EBADFD):
-                    raise e
-            self.socket = None
-            log.debug('Connection Closed: %s', self.id)
-
-        if self.switch and self.switch.connection is self:
-            self.switch.connection = None
-
-    def is_connected(self):
-        """Return True if it is connected.False otherwise."""
-        return self.socket is not None
-
-    def update_switch(self, switch):
-        """Update switch with this instance of Connection.
-
-        Parameters:
-          switch (:class:`~.core.switch.Switch`): switch instance.
-        """
-        self.switch = switch
-        self.switch.connection = self
 
 
 class Switch(object):
@@ -553,7 +445,7 @@ class Switch(object):
         return {'id': self.id,
                 'name': self.id,
                 'dpid': self.dpid,
-                'connection':  connection,
+                'connection': connection,
                 'ofp_version': self.ofp_version,
                 'type': 'switch',
                 'manufacturer': self.description.get('manufacturer', ''),
--- a/kytos/core/switch.py
+++ b/kytos/core/switch.py
@@ -1,7 +1,10 @@
 """Module with main classes related to Switches."""
 import json
 import logging
+from enum import Enum
+from errno import EBADFD, ENOTCONN
 from socket import error as SocketError
+from socket import SHUT_RDWR
 
 from pyof.v0x01.common.phy_port import PortFeatures
 
@@ -191,6 +194,25 @@ class Interface(object):
         return json.dumps(self.as_dict())
 
 
+class CONNECTION_STATE(Enum):
+    """Enum of possible general connections states."""
+
+    NEW = 0
+    SETUP = 1
+    ESTABLISHED = 2
+    FAILED = 3
+    FINISHED = 4
+
+
+class ProtocolInfo:
+    """Class to hold simple protocol information for the connection."""
+
+    def __init__(self):
+        self.name = None
+        self.version = None
+        self.state = None
+
+
 class Connection(object):
     """Connection class to abstract a network connections."""
 
@@ -207,6 +229,22 @@ class Connection(object):
         self.port = port
         self.socket = socket
         self.switch = switch
+        self.state = CONNECTION_STATE.NEW
+        self.protocol = ProtocolInfo()
+        self.remaining_data = b''
+
+    @property
+    def state(self):
+        """Return the state of the connection."""
+        return self._state
+
+    @state.setter
+    def state(self, new_state):
+        if new_state not in CONNECTION_STATE:
+            raise Exception('Unknown State', new_state)
+        self._state = new_state # noqa
+        log.info('Connection %s changed state: %s',
+                 self.id, self.state)
 
     @property
     def id(self):
@@ -226,14 +264,22 @@ class Connection(object):
         try:
             if self.is_connected():
                 self.socket.sendall(buffer)
-        except (OSError, SocketError):
+        except (OSError, SocketError) as exception:
+            log.debug('Could not send packet. Exception: %s', exception)
             self.close()
 
     def close(self):
         """Close the socket from connection instance."""
         if self.socket:
-            self.socket.close()
+            log.debug('Shutting down Connection %s', self.id)
+            try:
+                self.socket.shutdown(SHUT_RDWR)
+                self.socket.close()
+            except OSError as e:
+                if e.errno not in (ENOTCONN, EBADFD):
+                    raise e
             self.socket = None
+            log.debug('Connection Closed: %s', self.id)
 
         if self.switch and self.switch.connection is self:
             self.switch.connection = None
--- a/kytos/core/tcp_server.py
+++ b/kytos/core/tcp_server.py
@@ -4,36 +4,30 @@ from socket import error as SocketError
 from socketserver import BaseRequestHandler, TCPServer, ThreadingMixIn
 from threading import current_thread
 
-from pyof.v0x01.common.header import Header, Type
-
 from kytos.core.events import KytosEvent
-from kytos.core.switch import Connection
+from kytos.core.switch import CONNECTION_STATE, Connection
 
-__all__ = ('KytosServer', 'KytosOpenFlowRequestHandler')
+__all__ = ('KytosServer', 'KytosRequestHandler')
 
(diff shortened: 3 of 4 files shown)


erickvermot:
TODO: move constants to settings

kytos/core/switch.py
```diff
@@ -191,6 +192,24 @@ def as_json(self):
         return json.dumps(self.as_dict())
 
 
+class Connection_states(Enum):
```
diraol: `s/Connection_states/CONNECTION_STATE/`

Despite being a class, it will behave like a "constant", therefore I think we can use constants naming rules. @cemsbr Am I right here?
Or should we use `ConnectionState` instead?

kytos/core/switch.py
```diff
@@ -191,6 +192,24 @@ def as_json(self):
...
+    """Enum of possible general connections states."""
+
+    NEW = 0
+    SETUP = 1
+    ESTABLISHED = 2
+    FINISHED = 3
+
+
+class Protocol_info():
```
diraol: s/Protocol_info/Protocol/

(I was rewriting it to `ProtocolInfo`, but it is very difficult to distinguish the lowercase `L` and the uppercase `i`, so I think I just prefer `Protocol`. Or, eventually, `ProtocolData`, despite the fact that this is not a very good semantic name for its purpose.)

kytos/core/switch.py
```diff
@@ -207,6 +226,23 @@ def __init__(self, address, port, socket, switch=None):
         self.port = port
         self.socket = socket
         self.switch = switch
+        self._state = None
+        self.STATES = Connection_states
```
diraol: I don't like the definition of a `CONSTANT` inside the class as an attribute.

I really prefer using the `CONNECTION_STATE` class directly when needed. And also, I think that makes sense to define the default value from `state` as `CONNECTION_STATE.NEW` value.

Regarding the `@property` and the `setter`, we are following a "pythonic" approach on the project, which means no fixed types (as python) and no excessive checks like C++/Java would require. The user are responsible for what it is doing.
Being so, I think that we can just have the `state` attribute (and document it in a way that our users knows that this must be one option from `CONNECTION_STATE`).

diraol: But I am not absolutely sure about not using the `setter` here, just because of the "changed state logging". @beraldoleal and @cemsbr what do you think about it?

kytos/core/tcp_server.py
```diff
@@ -44,113 +40,108 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        # self.controller_put_raw_event = controller_put_raw_event
 
     def serve_forever(self, poll_interval=0.5):
-        """Handle requests until an explicit shutdown() request."""
+        """Handle requests until an explicit shutdown() is called."""
+        SOCKET_BIND_WAIT_TIME = 0.5
+        SOCKET_BIND_N_TRIES = 2
+        # Tries to bind the socket n times, it may still be
+        # in WAIT state from previous connection.
+        for i in range(SOCKET_BIND_N_TRIES):
```
diraol: @erickvermot take a look at the end of the section https://docs.python.org/3/library/socket.html#example. There it talks about `TIME_WAIT` on sockets, and the use of `socket.SO_REUSEADDR`. This may fix the problem you were trying to address here with this "loop" and it seems to be a more appropriate solution to the problem than an arbitrary loop.

kytos/core/tcp_server.py
```diff
@@ -44,113 +40,108 @@ def __init__(self, server_address, RequestHandlerClass,
...
     """
 
     def setup(self):
-        """Method used to create a new connection with client.
+        """Method used to setup the new connection.
 
-        This method builds a new connection, makes a new KytosEvent named
-        'kytos/core.connection.new' and put this in a raw buffer.
+        This method builds a new controller Connection, and places a
+        'kytos/core.connection.new' KytosEvent in the raw buffer.
```
diraol: `app` buffer

kytos/core/controller.py
```diff
@@ -355,10 +355,14 @@ def msg_out_event_handler(self):
 
             message = triggered_event.content['message']
             destination = triggered_event.destination
-            if destination:
+            if (destination and
+                    not destination.state == destination.STATES.FINISHED):
```
diraol: `destination.state == CONNECTION_STATE.FINISHED`

kytos/core/controller.py
```diff
@@ -355,10 +355,14 @@ def msg_out_event_handler(self):
...
                 destination.send(message.pack())
                 self.notify_listeners(triggered_event)
                 self.log.debug("MsgOutEvent handler called")
+            else:
+                self.log.debug(
+                    "connection is finished. Cannot send message")
```
diraol: `finished` or `closed`? What is the best "networking word" for this @beraldoleal

kytos/core/tcp_server.py
```diff
@@ -1,38 +1,34 @@
...
 
-    def __init__(self, server_address, RequestHandlerClass,
-                 # Change after definitions on #62
-                 # controller_put_raw_event):
-                 controller):
-        """Constructor of KytosServer receive the parameters below.
+    def __init__(self, server_address, RequestHandlerClass, controller):
+        """Constructor of KytosServer.
 
         Parameters:
```
diraol: @erickvermot as you are editing this file, could you please replace `Parameters` with `Args`? Tks!

kytos/core/tcp_server.py
```diff
@@ -44,113 +40,108 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        This method builds a new connection, makes a new KytosEvent named
-        'kytos/core.connection.new' and put this in a raw buffer.
+        This method builds a new controller Connection, and places a
+        'kytos/core.connection.new' KytosEvent in the raw buffer.
         """
         self.ip = self.client_address[0]
         self.port = self.client_address[1]
 
         self.connection = Connection(self.ip, self.port, self.request)
+        self.connection.state = self.connection.STATES.NEW
```
diraol: `self.connection.state = CONNECTION_STATE.NEW`

kytos/core/tcp_server.py
```diff
@@ -44,113 +40,108 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        When this method is called the request is closed, a new KytosEvent is
-        built with kytos/core.connection.lost name and put this in a app
-        buffer.
+        This method closes the connection socket and generates a
+        'kytos/core.connection.lost' KytosEvent in the App buffer.
         """
-        log.info("Client %s:%s disconnected. Reason: %s",
+        log.info("Connection lost with Client %s:%s. Reason: %s",
                  self.ip, self.port, self.exception)
+        self.connection.state = self.connection.STATES.FINISHED
```
diraol: `self.connection.state = CONNECTION_STATE.FINISHED`

kytos/core/tcp_server.py
```diff
@@ -44,113 +40,108 @@ def __init__(self, server_address, RequestHandlerClass,
...
+        self.request.shutdown(SHUT_WR)
```
diraol: @erickvermot what about the `graceful shutdown` method we discussed?

diraol (changes requested):

kytos/core/switch.py
erickvermot: I agree with the caps, I wrote with caps first but the changed it since it was a class...
I'll capitalize it back

kytos/core/switch.py
erickvermot: I think just ``Protocol`` is ok, specially if we add links to pack and unpack functions later.

kytos/core/switch.py
erickvermot: I don't see it as a definition, but simply as a shortcut so that when we receive the event as an arg on a handler, we can access the possible states without having to import something new... I'll remove it if you guys want.

About the setter, I thought it was a good way to provide the logging.

kytos/core/tcp_server.py
erickvermot: The example in the library forces the reuse of the address without waiting for the TIME_WAIT to expire. I think it would be better to wait for it before binding again.

kytos/core/tcp_server.py
erickvermot: I'm implementing it, but first I want to be sure where it makes sense to call the shutdown and where it makes sense to call close directly.

kytos/core/controller.py
erickvermot: It was meant to indicate the 'FINISHED' state, which does not necessarily mean the socket is actually closed.

kytos/core/tcp_server.py
diraol: Hum.... I still not sure if what way to go.
@beraldoleal ? @cemsbr ? @renanrodrigo ? @abaruchi ?
Any thoughts?
Just adding one more link:
http://www.unixguide.net/network/socketfaq/4.5.shtml

Just to clarify for those that are just getting into the discussion.

After a socket is closed, it's state is changed to "closed" but it remains opened for some time (`TIME_WAIT`). This is a default behavior from Unix sockets.
If anyone tries to establish a new socket on the same "socket"*, then the OS will raise `Address already in use` error.

We have three options here.
1) Just raise the error;
2) Use a `timeout` and keep trying to bind to the socket again. [ @erickvermot proposal on this PR ];
3) Use the `socket.SO_REUSEADDR` to tell the OS that it should get the living (but closed) socket and "restart" it.

*A socket is a 5 tuple (`proto`, `local addr`, `local port`, `remote addr`, `remote port`)

renanrodrigo: I agree with @erickvermot on option 2.
`socket.SO_REUSEADDR` could cause problems in case of quick switch reconnections if they are exchanging data with the controller.

kytos/core/switch.py
diraol: If we are going to use the setter here, the best approach would be something like this:
```
class Test():
    def __init__(self):
        self.x = None
    @property
    def x(self):
        return self.__x
    @x.setter
    def x(self, x):
        self.__x = x
        print('setting x with {}'.format(x))
```

kytos/core/tcp_server.py
diraol: We ( @erickvermot , @abaruchi , @renanrodrigo and I) discussed this and the possibility of errors due to reusing the connection, considering the OpenFlow protocol (or eventually any other SDN protocol) is negligible. Even if it occurs, the consequences are minimal and manageable, in a way that the costs of waiting the `TIME_WAIT` to expire seems to be much higher than just reusing the socket and eventually dropping some lost packages or restarting the connection to a switch.

So, let's stick with the option 3 and reuse the socket without waiting the `TIME_WAIT`.

erickvermot: I still believe there should be at least an option to deal with that like 'FORCEBIND' or similar. I'm not sure the consequences are negligible, and the default behaviour of TCP servers is to wait.
Since the others prefer to go with 3 and ignore the possible problems, I'll change the PR.

diraol (commented):
There is still the need to set the `socket.SO_REUSEADDR` flag on the socket. Have you done it? I haven't seen it...

erickvermot:
well, seen or not, it's there...

kytos/core/tcp_server.py
erickvermot: @abaruchi @renanrodrigo @diraol @beraldoleal,
I've moved my solution to another branch in case someone wants to consider it: https://github.com/erickvermot/kytos/tree/socket_bind_fix_wait
It retries N times waiting X seconds between retries, and only forces the bind if the retries fails.
This gives at least some time (N*X) before forcing the bind, and may help prevent packets from an old connection and possibly from another switch to eventually be received by the new connection.

kytos/core/controller.py
beraldoleal: Let's use "Connection closed" because is an useful info to the network administrator. If we need more details like "FYN_WAIT" or something use as debug message.

kytos/core/switch.py
```diff
@@ -191,6 +193,24 @@ def as_json(self):
...
+class CONNECTION_STATE(Enum):
+    """Enum of possible general connections states."""
+
+    NEW = 0
+    SETUP = 1
+    ESTABLISHED = 2
+    FINISHED = 3
+
+
+class Protocol():
```
beraldoleal: `class Protocol:`

beraldoleal: s/Protocol/ApplicationProtocol/g

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        This method builds a new connection, makes a new KytosEvent named
-        'kytos/core.connection.new' and put this in a raw buffer.
+        This method builds a new controller Connection, and places a
+        'kytos/core.connection.new' KytosEvent in the app buffer.
         """
         self.ip = self.client_address[0]
         self.port = self.client_address[1]
 
         self.connection = Connection(self.ip, self.port, self.request)
+        self.connection.state = CONNECTION_STATE.NEW
```
beraldoleal: Make `CONNECTION_STATE.NEW` as default. If yes, please remove this line.

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
         self.request.settimeout(30)
         self.exception = None
 
         event = KytosEvent(name='kytos/core.connection.new',
```
beraldoleal: Please, emit the event according to the port. Ex: If port is 6633 the event should be `kytos/core.connection.new.openflow`

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
             try:
-                while len(raw_header) < header_len:
-                    remaining = header_len - len(raw_header)
-                    raw_header += self.request.recv(remaining)
-                    if not raw_header:
-                        break
-            except InterruptedError as exception:
+                new_data = self.request.recv(MAX_SIZE)
+            except (SocketError, OSError, InterruptedError) as exception:
                 self.exception = exception
```
beraldoleal: s/exception/last_exception/g

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
+                log.debug('Socket handler exception while reading: %s',
+                          exception)
+                break
+            if new_data == b'':
+                self.exception = 'Null Request (closed by client)'
```
beraldoleal: "Request closed by client"

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
-                self.exception = exception
-                break
-
             content = {'source': self.connection,
-                       'header': header,
-                       'binary_data': binary_data}
-
-            event = KytosEvent(name='kytos/core.messages.openflow.new',
+                       'new_data': new_data}
+            event = KytosEvent(name='kytos/core.server.data.in',
```
beraldoleal: Here the same, add the port number on the event.

kytos/core/tcp_server.py
```diff
@@ -44,113 +38,93 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        built with kytos/core.connection.lost name and put this in a app
-        buffer.
+        This method closes the connection socket and generates a
+        'kytos/core.connection.lost' KytosEvent in the App buffer.
         """
-        log.info("Client %s:%s disconnected. Reason: %s",
+        log.info("Connection lost with Client %s:%s. Reason: %s",
                  self.ip, self.port, self.exception)
+        self.connection.state = CONNECTION_STATE.FINISHED
+        self.request.shutdown(SHUT_RDWR)
```
beraldoleal: Please do a double check if is really necessary close after shutdown.

kytos/core/tcp_server.py
beraldoleal: `kytos/core.(protocol|port).data.in`

beraldoleal (changes requested):

erickvermot:
I have updated the branch, but I'm experiencing some problems when I test it with mininet.
I do not have any problems when testing with a generic fake switch I made, but mininet seems to be sending many 'packet-in' to the controller, even though I'm not telling him to ping or anything.
I'm still investigating what is happening. 

\+ test is complaining that:
kytos/core/tcp_server.py:1:1: E0203 Access to member 'connection' before its definition line 82 [pylint]
but I believe the 'access' it it worrying about is the one in switch.py line 426 where we check if the connection exists...

kytos/core/switch.py
```diff
@@ -207,6 +229,22 @@ def __init__(self, address, port, socket, switch=None):
...
+        """Return the state of the connection."""
+        return self._state
+
+    @state.setter
+    def state(self, new_state):
+        if new_state not in CONNECTION_STATE:
+            raise Exception('Unknown State', new_state)
+        self._state = new_state # noqa
+        log.info('Connection %s changed state: %s',
+                  self.id, self.state)
```
diraol: There is an indentation error here (1 extra space on line 247)

kytos/core/tcp_server.py
```diff
@@ -44,118 +37,112 @@ def __init__(self, server_address, RequestHandlerClass,
...
-        'kytos/core.connection.new' and put this in a raw buffer.
+        This method builds a new controller Connection, and places a
+        'kytos/core.connection.new' KytosEvent in the app buffer.
         """
         self.ip = self.client_address[0]
         self.port = self.client_address[1]
 
+        log.info("New connection from %s:%s", self.ip, self.port)
+
         self.connection = Connection(self.ip, self.port, self.request)
```
diraol: Add a `# noqa` on the end of this line to avoid `pylint` error `E0203 Access to member 'connection' before its definition`.

diraol (changes requested):

kytos/core/tcp_server.py
erickvermot: It only catches one exception, should I change it all the same?

kytos/core/tcp_server.py
erickvermot: adding #noqa on line 82 did not solve it, since tox complains on tcp_server line 1 (probably when importing..)

diraol: O_o That's really strange..... I'll run it locally to see what is going on....

erickvermot:
@beraldoleal ping

beraldoleal:
@erickvermot the problems with mininet that you said also happen without this PR ?

Also you have to fix the pylint errors.

erickvermot:
@beraldoleal, about mininet, it was no problem at all, just lldp doing its job. But yes, it happens with or without this PR.

About the pylint errors, I talked about this a few weeks ago with @diraol, https://github.com/kytos/kytos/pull/374#issuecomment-298771159 and he also doesn't know how to fix it or how to make it be ignored... Any suggestion?
kytos/core/tcp_server.py:1:1: E0203 Access to member 'connection' before its definition line 82 [pylint]

About the commits, would you like me to combine the commits to a single one to make it more clean?

kytos/core/tcp_server.py
```diff
@@ -1,40 +1,33 @@
...
 
 class KytosServer(ThreadingMixIn, TCPServer):
-    """Abstraction a TCPServer to listen packages from the network.
+    """Abstraction of a TCPServer to listen to packages from the network.
 
-    This is a TCP Server that will be listening on the specific port
-    (defaults to 6633) for any package that comes from the network and then
-    stabilishes the socket connection to the devices (switches) when needed,
-    considering a thread per connection and keeping the connection alive.
+    The KytosServer will listen on the specified port (defaults to 6633)
```
beraldoleal: Please, remove references to 6633 on this comments.

kytos/core/switch.py
```diff
@@ -191,6 +194,25 @@ def as_json(self):
...
+    SETUP = 1
+    ESTABLISHED = 2
+    FAILED = 3
+    FINISHED = 4
+
+
+class ProtocolInfo:
+    """Class to hold simple protocol information for the connection."""
+
+    def __init__(self):
```
beraldoleal:
```
def __init__(self, name = None, version = None, state = None):
   self.name = name,
   ...
```

kytos/core/switch.py
```diff
@@ -191,6 +194,25 @@ def as_json(self):
...
+    """Enum of possible general connections states."""
+
+    NEW = 0
+    SETUP = 1
+    ESTABLISHED = 2
+    FAILED = 3
+    FINISHED = 4
+
+
+class ProtocolInfo:
```
beraldoleal: s/ProtocolInfo/ConnectionProtocol/g

kytos/core/switch.py
```diff
@@ -191,6 +194,25 @@ def as_json(self):
         return json.dumps(self.as_dict())
 
 
+class CONNECTION_STATE(Enum):
```
beraldoleal: move `CONNECTION_STATE`, `ConnectionProtocol` and `Connection` to a new module , named `connection`

beraldoleal (changes requested):
One more time, please. This is the last one. :)

PR #375, opened 2017-04-25
Wheel generation

cemsbr:
`python setup.py bdist_wheel` generates dist/*.whl

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ipython
 lockfile
-python-openflow
+python-openflow==2017.1b1.dev1
 flask-socketio
 flask
 python-daemon>=2.1.2
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,6 @@ import sys
 from abc import abstractmethod
 # Disabling checks due to https://github.com/PyCQA/pylint/issues/73
 from distutils.command.clean import clean  # pylint: disable=E0401,E0611
-from pathlib import Path
 from subprocess import call
 
 from setuptools import Command, find_packages, setup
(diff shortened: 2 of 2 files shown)


cemsbr:
@renanrodrigo, I certified that setup.py install and bdis_wheel are working in a clean virtualenv.

renanrodrigo:
@cemsbr Nice! WIll run all tests and post a report soon

requirements.txt
```diff
@@ -1,6 +1,6 @@
 ipython
 lockfile
-python-openflow
+python-openflow==2017.1b1.dev1
```
diraol, line 3: @cemsbr why not `>=` ?

cemsbr: Because it wil install 2017.1b1. This is because the version string is wrong. It should be either 2017.1b1.post1 or 2017.b2.dev1. @beraldoleal, do you want to correct this or we just pay more attention to next releases?

setup.py
```diff
@@ -187,16 +179,9 @@ def generate_file_link(file_name):
...
-    def create_paths():
-        """Method used to create the paths used by Kytos in develop mode."""
-        directories = [os.path.join(BASE_ENV, 'etc/kytos')]
-        for directory in directories:
-            if not os.path.exists(directory):
-                os.makedirs(directory)
-
 
-requirements = [i.strip() for i in open("requirements.txt").readlines()]
+setup_requires = ['jinja2']
```
beraldoleal, line 183: I'm going to accept this, because I need to see the things moving on. But this must be fixed.

PR #376, opened 2017-04-26
Improving NappManager to install napp dependencies

macartur:


kytos/core/napps/manager.py
```diff
@@ -95,6 +95,12 @@ def install(self, napp_uri, enable=True):
 
         log.info("New NApp installed: %s", napp)
 
+        napp = NApp.create_from_json(dst/'kytos.json')
+        log.info('Installing Napp Dependencies:')
+        if napp.napp_dependencies:
```
diraol: I think we do not need this "if".
The `napp_dependencies` attribute must always exist. It should always be a list. If there are no dependencies, it should be an empty list.
Therefore, the `for napp in napp.napp_dependencies` will not fail it `napp_dependencies` is an empty list.

diraol (changes requested):

diraol (approved):

kytos/core/napps/manager.py
```diff
@@ -79,8 +79,8 @@ def install(self, napp_uri, enable=True):
             return False
 
         if not napp.repository:
-            # Check here what is configured repo
-            pass
+            options = self._controller.options
+            napp.repository = json.loads(options.napps_repositories)[0]
```
cemsbr: I think it would be better to set the options.napps_repositories as an array where the config file is read. After that, you will also have to change a line in manager.py

kytos/core/napps/manager.py
```diff
@@ -95,6 +95,11 @@ def install(self, napp_uri, enable=True):
 
         log.info("New NApp installed: %s", napp)
 
+        napp = NApp.create_from_json(dst/'kytos.json')
+        log.info('Installing Napp Dependencies:')
```
cemsbr: Checking whether there're deps to be installed and fixing caps in log message:
```python
if napp.napp_dependencies:
    log.info('Installing NApp dependencies:')
```

cemsbr (changes requested):
Please, add a new commit instead of amending.

kytos/core/napps/manager.py
macartur: When I tried to execute the self._controller.options the result is a json, not an array. I will try to change this line.

kytos/core/napps/manager.py
macartur: OK.

cemsbr (approved):

kytos/core/napps/manager.py
```diff
@@ -95,6 +94,11 @@ def install(self, napp_uri, enable=True):
 
         log.info("New NApp installed: %s", napp)
 
+        napp = NApp.create_from_json(dst/'kytos.json')
+        log.info('Installing Napp dependencies:')
+        for napp in napp.napp_dependencies:
+            self.install(napp, enable)
```
diraol: The `install()` method receives, as argument, a `napp_uri` and, optionally, a `enable` (boolean) arg.

Here you are passing `napp` (NApp object) instead of the `napp_uri`. Is this right?

diraol (changes requested):

kytos/core/napps/manager.py
diraol: I  think that @cemsbr suggested to put the "log.info" about "installing dependencies" inside an if clause, since that if there are no dependencies there is no reason to log this message.

beraldoleal:
@diraol your suggestions are done ? Are you ok with this PR ?

kytos/core/napps/manager.py
```diff
@@ -95,6 +94,11 @@ def install(self, napp_uri, enable=True):
 
         log.info("New NApp installed: %s", napp)
 
+        napp = NApp.create_from_json(dst/'kytos.json')
+        log.info('Installing Napp dependencies:')
+        for napp_dependency_uri in napp.napp_dependencies:
+            self.install(napp_dependency_uri, enable)
+
```
diraol: @macartur, change this block to:
```
napp = NApp.create_from_json(dst/'kytos.json')
if napp.napp_dependencies:
    if len(napp.napp_dependencies) == 1:
        log.info('Installing NApp %s dependency', <NAPP_ID>)
        self.install(napp_dependency_uri, enable)
    else:
        log.info('Installing NApp %s dependencies', <NAPP_ID>)
        for napp_dependency_uri in napp.napp_dependencies:
            self.install(napp_dependency_uri, enable)
```

diraol (changes requested):

kytos/core/napps/manager.py
macartur: Fixed.

beraldoleal: Why a if else here ? `napp.napp_dependencies` is a list right ? 

There is no problem if there is no dependency and a simple print is output
```
napp = NApp.create_from_json(dst/'kytos.json')
log.info('Installing NApp %s dependencies', <NAPP_ID>)
for napp_dependency_uri in napp.napp_dependencies:
        self.install(napp_dependency_uri, enable)
```

diraol (approved):

PR #377, opened 2017-04-26
Improving logs

macartur:
Set logging configuration using  etc/kytos/logging.ini
Fixed syslog

### DIFF
--- a/etc/kytos/logging.ini
+++ b/etc/kytos/logging.ini
@@ -1,38 +1,38 @@
 [formatters]
-keys: precise,brief
- 
+keys: console,syslog
+
 [handlers]
 keys: console,syslog
- 
+
 [loggers]
-keys: root,daemon,cli
- 
-[formatter_brief]
-format: %(name)s:%(levelname)s:  %(message)s
- 
-[formatter_precise]
+keys: root,api_server,socket
+
+[formatter_syslog]
 format: %(name)s:%(levelname)s %(module)s:%(lineno)d:  %(message)s
- 
+
+[formatter_console]
+format: %(asctime)s - %(levelname)s [%(name)s] (%(threadName)s) %(message)s
+
 [handler_console]
 class: StreamHandler
-args: []
-formatter: brief
- 
+args:[sys.stdout]
+formatter: console
+
 [handler_syslog]
 class: handlers.SysLogHandler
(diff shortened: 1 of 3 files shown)


etc/kytos/logging.ini
```diff
@@ -1,38 +1,46 @@
...
-args: []
-formatter: brief
- 
+args:[sys.stdout]
+formatter: thread
+
 [handler_syslog]
 class: handlers.SysLogHandler
-args: [('localhost', handlers.SYSLOG_UDP_PORT), handlers.SysLogHandler.LOG_USER]
+args: ['/dev/log']
```
diraol: @macartur this should be `/var/log` doesn't it?

kytos/core/logs.py
```diff
@@ -27,14 +27,14 @@ def add_stream_handler(cls, stream):
...
-        handler = logging.StreamHandler(stream)
+        handler = StreamHandler(stream)
         cls._add_handler(handler)
         return handler
 
     @staticmethod
     def _add_handler(handler):
-        handler.setFormatter(logging.Formatter(FORMAT))
-        logging.getLogger().addHandler(handler)
+        handler.setFormatter(Formatter(FORMAT))
```
diraol: Can't you retrieve the format from one of those defined on the `logging.ini` file?

diraol (changes requested):

etc/kytos/logging.ini
macartur: This is the device used to write in local syslog. I will try to change this.

macartur: I am using that http://scottbarnham.com/blog/2008/01/01/sysloghandler-not-writing-to-syslog-with-python-logging/, I can use syslog using logging.ini yet.

cemsbr: I just noticed that I have /dev/log :). I did a quick read in the link above and also in Python's docs and I thinks it's ok.

diraol: Agreed!

diraol:
@macartur also do not forget to mention which issue you are dealing with on you commit(s) message(s) =)

beraldoleal:
@macartur waiting for reply on @diraol comment.

diraol (approved):

Issue #381, opened 2017-05-03
Kytos execution as normal user, and PIDfile creation


erickvermot:
Since PR https://github.com/kytos/kytos/pull/360 It has not been possible anymore to run kytos as a normal user after installing it on the system as root.
Issue https://github.com/kytos/kytos/issues/368 was opened about this, and then closed with https://github.com/kytos/kytos/pull/369, but the problem remains and It is not possible to run kytos as a normal user.
Furthermore, if one installs kytos using setup.py, a /var/run/kytos folder is created (with permission 777) so that a pidfile can be written inside it during execution, but the problem is that (on Arch which uses systemd) this folder is removed after reboot. This makes kytosd unable to run even by root, and complains about permission denied to write to the file, simply because the folder does not exists.
@renanrodrigo @diraol @cemsbr @beraldoleal

erickvermot:
https://askubuntu.com/questions/303120/how-folders-created-in-var-run-on-each-reboot talks about /var/run and systemd

erickvermot:
![2017-05-03-170818_1337x559_scrot](https://cloud.githubusercontent.com/assets/10514972/25679276/34851720-3023-11e7-89ac-2a50a0e5d460.png)

erickvermot:
since https://github.com/kytos/kytos/pull/387 it has been possible to run kytosd as normal user after install, but this is no longuer possible after reboot, since /var/run/kytos is removed and a normal user cannot create it anymore.
If one runs it using root, the deamon starts ok, but then a normal user cannot run it because the pid file and folder belongs to root and has permission 755 (probably because of the umask).

simple workaround is to create the folder with permissive permissions: ``sudo mkdir -m 777 /var/run/kytos`` and then you can start kytosd as a normal user.

cemsbr:
This is already solved. Reopen if you still have this problem.

erickvermot:
@cemsbr, I do not understand why you closed this.
Could you show me which action was taken to solve this?
The problem still stands, and I haven't seen any PR related to this since my comment, nor any reason to close it.
@beraldoleal may I reopen this?

cemsbr:
I closed because it was fixed in another issue. Did you erase /var/run/kytos folder before running the latest version?

erickvermot:
oh, which issue? 
Problem still stands for me...
I do not understand your point about erasing /var/run/kytos, since the problem exists around the fact that after reboot /var/run/kytos does not exist anymore, and a regular user cannot create it..
```
[...]

PR #382, opened 2017-05-04
Create folder for pidfile in /var/run when running

renanrodrigo:
This is a workaround for #381. It does not allow kytos to be run as a regular user, so the issue shall stay open.
This fixes the problem of running as root after a restart.
The permanent solution to #381 needs further discussion.

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -118,6 +118,9 @@ class Controller(object):
         self.enable_logs()
         pid = os.getpid()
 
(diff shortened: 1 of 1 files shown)


diraol (approved):

cemsbr (changes requested):
No permission for regular user.

cemsbr (commented):
Wondering about permissions

cemsbr (approved):

PR #383, opened 2017-05-04
web-ui CSS Compile - FIX #380

### LINKED ISSUE #380: Pre-compile web-ui during install
We need to compile the CSS for the web-ui during the install procedure.

diraol:


diraol:
@beraldoleal I changed the PR to build the css while building the package/installing.
For now we are not using the minified version of the css in order to avoid putting one more minified file on the package (wheel).

PR #387, opened 2017-05-04
Run kytos installed by root as regular user

cemsbr:
- Set permission of /var/run/kytos like /tmp
- Move pid file deletion from controller.stop() to Python's atexit
- Change level from info to error if sys.exit is to be called
- Update some comments and error messages


cemsbr:
Tip: an easy way to test: pip install git+https://github.com/cemsbr/kytos@pid_file

cemsbr:
@erickvermot, we were discussing about this today.

PR #401, opened 2017-05-08
Fix path to pidfile in config.py

### LINKED ISSUE #398: Wrong pid file location on config.py
Setup.py creates the folder ``/var/run/kytos`` to hold the pidfile, but config.py is still referencing ``/var/run/kytosd.pid``

renanrodrigo:
Fix #398


beraldoleal:
@renanrodrigo tests...

PR #402, opened 2017-05-08
fixes BASE_ENV prefix used on setup for conf template

erickvermot:


erickvermot:
#403 is better. I'm closing this.

PR #403, opened 2017-05-08
Fixing template prefix when BASE_ENV is '/'

macartur:


beraldoleal:
Duplicated of #402 ?

erickvermot:
@macartur is right, #402 still outputs double slashes when BASE_ENV is '/'. I'll close #402, this is a better solution.

diraol (approved):

etc/kytos/kytos.conf.template
```diff
@@ -21,15 +21,15 @@ daemon = False
...
+logging = {{ prefix }}etc/kytos/logging.ini
 
 listen = 0.0.0.0
 
 port = 6633
 
 # Where should the controller look for network apps ?
 # This directory has both core napps and user installed napps.
-napps = {{ prefix }}/var/lib/kytos/napps
+napps = {{ prefix }}var/lib/kytos/napps
```
beraldoleal: When we read things like this looks odd. I prefer `{{ prefix }}/var/lib/kytos/napps`. In order to do this, we have to make sure that prefix is something without a ending slash or something empty.

So if prefix == / lets make prefix == "". And additionally read my next comment.

setup.py
```diff
@@ -128,6 +128,9 @@ def generate_file_from_template(cls, templates,
         """
         from jinja2 import Template
 
+        if not kwargs.get('prefix').endswith('/'):
+            kwargs['prefix'] += '/'
+
```
beraldoleal: Based on the previous comment, lets do the oposite. If prefix ends with '/' lets remove the last '/'.

cemsbr (approved):

beraldoleal (changes requested):

beraldoleal:
@macartur ?

macartur:
@beraldoleal  Fixed.

PR #410, opened 2017-05-09
Catch exception ConnectionResetError

macartur:
This commit will avoid thrown exception ConnectionResetError when the switch
is reconnecting.


diraol (commented):
@beraldoleal When a connection is resetted by the peer (ConnectionResetError) should we "close the socket", as if it was disconnected or should we start a procedure to bring that connection to the initial state (without handshake, for example)?

beraldoleal:
@diraol we have to discuss more on this, but I think so. Yes we should.

I will accept this PR, but you can open another issue just to fix this "initial state problem"

PR #418, opened 2017-05-11
Improving log.py to work when syslog is down

### LINKED ISSUE #400: kytosd needs syslog up and running
If syslog is not running, kytosd fails...

macartur:
Fix #400


diraol (approved):
Seems to do what was requested on the issue, thus closing it.

PR #422, opened 2017-05-15
Improving APIServer to use kytos.conf file

macartur:


diraol (changes requested):
@macartur on  the `APIServer` class you have added, on the `__init__`, and argument `options`, and the description is:

> options(Namespace): Namespace used to configure the api server.

But you are passing the `Controller` options, not a section specific for the `api_server`. Which means that `options.debug` is not a `api_server` specific configuration, but a configuration for the whole controller. And the same goes for `options.listen` and `options.port`. This can cause confusion between the TCPServer listen address/port and the APIServer listen address/port.
I think that we need a more error prone solution over here, either using more specific/descriptive variables names, either using 'sections' on the config file.

Besides that, the `set_debug` method is looking a bit strange to me.
If I call `self.set_debug('WARNING')`, then the logger level will be set to "DEBUG" instead of WARNING.
This behavior was already there, but, since you are editing the method, I think we can try to find out a better solution. May be change the method name to `set_log_level` (making the default to be warning), or something else. Feel free to think about a solution and propose it. =)

diraol:
@macartur you haven't  indicated on the commits or on the PR message which issue is being handled on this PR. Do not forget to add it in order to close the issue (if it should be closed) when the PR is accepted.

beraldoleal:
@diraol are you ok with this PR ?

diraol (approved):

diraol:
@beraldoleal yes I am, @macartur fix the conflicts, please ;)

diraol:
And remember to update the `CHANGELOG` if it make sense.

beraldoleal:
@macartur scrutinizer.

beraldoleal:
Hi @macartur thanks so much, nice work.

I know that this was not introduced by this PR. But can you please describe better listen, port, api_port on kytos.conf.template ? (as comments)

PR #426, opened 2017-05-16
Catching SystemExit before start the kytos' shell

### LINKED ISSUE #408: The call sys.exit() is not working in Kytos
When we have the call sys.exit(), the exception below is thrown and Kytos keep running.

```
/home/macartur/Desktop/unesp/kytos/kytos/kytos/core/controller.py in create_pidfile(self)
    154                 self.log.error("PID file %s exists. Delete it if Kytos is not"   
    155                                "running. Aborting.")                             
--> 156                 sys.exit(os.EX_CANTCREAT)                                        
    157             except OSError:                                                      
    158                 try:                                                             
                                                                                         
SystemExit: 73                                                                           
```

macartur:
Avoid start Kytos' Shell when thrown exception while starting the controller.

Fix #408

### DIFF
--- a/bin/kytosd
+++ b/bin/kytosd
@@ -3,6 +3,7 @@
 import daemon
 import IPython
 import signal
+import sys
 
 from IPython.terminal.embed import InteractiveShellEmbed
 from IPython.terminal.prompts import Prompts, Token
@@ -55,7 +56,14 @@ if __name__ == '__main__':
     config = KytosConfig()
     controller = Controller(config.options['daemon'])
     if controller.options.foreground:
-        controller.start()
+
+        try:
+            controller.start()
+        except SystemExit as e:
+            controller.log.error(e)
+            controller.log.info("Kytos start aborted.")
+            sys.exit()
+
         address = controller.server.server_address[0]
         port = controller.server.server_address[1]
         banner1 += " tcp://{}:{}\n".format(address, port)
@@ -73,5 +81,9 @@ if __name__ == '__main__':
         controller.stop()
     else:
         with daemon.DaemonContext():
-            controller.start()
-        print('Kytos daemon started.')
+            try:
+                controller.start()
+            except SystemExit as e:
+                controller.log.error(e)
+                controller.log.info("Kytos daemon start aborted.")
+                sys.exit()
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -151,16 +151,15 @@ class Controller(object):
                 # If kill() doesn't return an error, there's a process running
                 # with the same PID. We assume it is Kytos and quit.
                 # Otherwise, overwrite the file and proceed.
-                self.log.error("PID file %s exists. Delete it if Kytos is not"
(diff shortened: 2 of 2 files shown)


bin/kytosd
```diff
@@ -55,7 +56,12 @@ if __name__ == '__main__':
     config = KytosConfig()
     controller = Controller(config.options['daemon'])
     if controller.options.foreground:
-        controller.start()
+
+        try:
+            controller.start()
+        except SystemExit:
+            sys.exit()
```
beraldoleal: Please print some useful information when exiting... for instance the reason of SystemExit exception.

beraldoleal (changes requested):

diraol (approved):

bin/kytosd
```diff
@@ -73,5 +81,10 @@ if __name__ == '__main__':
         controller.stop()
     else:
         with daemon.DaemonContext():
-            controller.start()
-        print('Kytos daemon started.')
+            try:
+                controller.start()
+                print('Kytos daemon started.')
```
beraldoleal: This should be a log right ?

bin/kytosd
```diff
@@ -55,7 +56,14 @@ if __name__ == '__main__':
     config = KytosConfig()
     controller = Controller(config.options['daemon'])
     if controller.options.foreground:
-        controller.start()
+
+        try:
+            controller.start()
+        except SystemExit as e:
+            print(e)
+            print("Kytos start aborted.")
```
beraldoleal: Can this be a log ?

kytos/core/controller.py
```diff
@@ -151,16 +151,15 @@ def create_pidfile(self):
                 # If kill() doesn't return an error, there's a process running
                 # with the same PID. We assume it is Kytos and quit.
                 # Otherwise, overwrite the file and proceed.
-                self.log.error("PID file %s exists. Delete it if Kytos is not"
-                               "running. Aborting.")
-                sys.exit(os.EX_CANTCREAT)
+                error_msg = ("PID file {} exists. Delete it if Kytos is not "
```
beraldoleal, line 154: Why did you removed self.log.error ? This should be logged too.

kytos/core/controller.py
```diff
@@ -207,10 +206,8 @@ def start_controller(self):
             try:
                 thread.start()
             except OSError as e:
-                self.log.error("Error starting thread %s", thread)
-                self.log.error(e)
-                self.log.error("Kytos start aborted.")
-                sys.exit()
```
beraldoleal, line 213: Here too

beraldoleal (changes requested):

bin/kytosd
macartur: Yes, I will change this.

kytos/core/controller.py
macartur: This message is logged too, when the SystemExit is raised the error message is passed and the controller will display this in the log.

PR #435, opened 2017-05-22
Refactor Flow.from_dict method in flow.py

renanrodrigo:
The method was simplified to receive only a flow dict, instead of a "dict within a dict". I checked and found no occurrences of this method outside the Flow Manager NApp.

### DIFF
--- a/kytos/core/flow.py
+++ b/kytos/core/flow.py
@@ -151,17 +151,17 @@ class Flow(object):
         Returns:
             flow (:class:`~kytos.core.flow.Flow`): Flow built from json.
         """
-        flow_fields = dict_content.get('flow', {})
         flow = Flow()
 
-        for attribute_name, value in flow_fields.items():
+        for attribute_name, value in dict_content.items():
+            # "actions" key has a list of actions as value, and each action on
+            # the list must be created separately
             if attribute_name == "actions":
-                continue
-            setattr(flow, attribute_name, value)
-
-        for dict_action in flow_fields.get('actions', []):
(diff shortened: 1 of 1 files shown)


kytos/core/flow.py
```diff
@@ -151,15 +151,14 @@ def from_dict(dict_content):
         Returns:
             flow (:class:`~kytos.core.flow.Flow`): Flow built from json.
         """
-        flow_fields = dict_content.get('flow', {})
         flow = Flow()
 
-        for attribute_name, value in flow_fields.items():
+        for attribute_name, value in dict_content.items():
             if attribute_name == "actions":
```
diraol: @renanrodrigo this is quite obvious for now, but could you please add a comment over here on why you are 'skipping' the `actions` attribute?

Besides that, two suggestions of different approaches:
1)
```python
flow = Flow()

for action_dict in dict_content.pop('actions', []):
    action = OutputAction.from_dict(action_dict)
    flow.actions.append(action)

for attribute_name, value in dict_content.items():
    setattr(flow, attribute_name, value)
```
2)
```python
flow = Flow()

for attribute_name, value in dict_content.items():
    if attribute_name == 'action':
        for action_dict in value:
            action = OutputAction.from_dict(action_dict)
            flow.actions.append(action)
    else:
        setattr(flow, attribute_name, value)
```

diraol (changes requested):

renanrodrigo:
@diraol done!

diraol (approved):

PR #439, opened 2017-05-30
Add endpoint to display controller.options

macartur:
Change the class controller to register all endpoints before start the APIServer.
Add the endpoint '/kytos/configuration' to render the options loaded by kytos.
Add some test for kytos/core/{controller, api_server}.py
Fix kytos/kytos-utils#86

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ All notable changes to the kytos project will be documented in this file.
 ********************************
 Added
 =====
+- '/kytos/configuration/' endpoint to display kytos configuration.
 
 Changed
 =======
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -83,30 +83,29 @@ class APIServer:
         """Return string with routes registered by Api Server."""
         return [x.rule for x in self.app.url_map.iter_rules()]
 
-    def register_kytos_routes(self):
+    def register_api_server_routes(self):
         """Register initial routes from kytos using ApiServer.
 
-        Initial routes are: ['/kytos/status', '/kytos/shutdown']
+        Initial routes are: ['/kytos/status/', '/kytos/shutdown/']
         """
         if '/kytos/status/' not in self.rest_endpoints:
-            self.app.add_url_rule('/kytos/status/', self.status_api.__name__,
-                                  self.status_api, methods=['GET'])
+            self.register_rest_endpoint('/status/',
+                                        self.status_api, methods=['GET'])
 
         if '/kytos/shutdown' not in self.rest_endpoints:
-            self.app.add_url_rule('/kytos/shutdown',
-                                  self.shutdown_api.__name__,
-                                  self.shutdown_api, methods=['GET'])
+            self.register_rest_endpoint('/shutdown/',
+                                        self.shutdown_api, methods=['GET'])
 
     @staticmethod
     def status_api():
-        """Display json with kytos status using the route '/kytos/status'."""
+        """Display json with kytos status using the route '/kytos/status/'."""
         return '{"response": "running"}', 201
 
     @staticmethod
     def stop_api_server():
         """Method used to send a shutdown request to stop Api Server."""
         try:
-            urlopen('http://127.0.0.1:8181/kytos/shutdown')
+            urlopen('http://127.0.0.1:8181/kytos/shutdown/')
         except URLError:
             pass
 
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -14,6 +14,7 @@ Basic usage:
     controller.start()
 """
 import atexit
+import json
 import logging
 import os
 import re
@@ -89,13 +90,19 @@ class Controller(object):
         #: The key is the switch dpid, while the value is a Switch object.
         self.switches = {}  # dpid: Switch()
 
+        #: datetime.datetime: Time when the controller finished starting.
         self.started_at = None
 
+        #: dict: Hash with all websockets used by Kytos.
         self.websockets = {}
 
+        #: logging.Logger: Logger instance used by Kytos.
         self.log = None
 
+        #: API Server used to expose rest endpoints.
         self.api_server = APIServer(__name__, self.options.debug)
+
+        self.register_kytos_endpoints()
 
         #: Adding the napps 'enabled' directory into the PATH
         #: Now you can access the enabled napps with:
@@ -172,8 +179,6 @@ class Controller(object):
         Starts a thread for each buffer handler.
         Load the installed apps.
         """
-        self.api_server.register_kytos_routes()
-        self.api_server.register_web_ui()
         self.register_websockets()
         self.log.info("Starting Kytos - Kytos Controller")
         self.server = KytosServer((self.options.listen,
@@ -213,9 +218,30 @@ class Controller(object):
         self.load_napps()
         self.started_at = now()
 
+    def register_kytos_endpoints(self):
+        """Register all rest endpoint served by kytos.
+
+        -   Register APIServer endpoints
+        -   Register WebUI endpoints
+        -   Register '/kytos/config' endpoint
+        """
+        self.api_server.register_api_server_routes()
+        self.api_server.register_web_ui()
+        self.api_server.register_rest_endpoint('/config/',
+                                               self.configuration_endpoint,
+                                               methods=['GET'])
+
     def register_rest_endpoint(self, *options, **kwargs):
         """Method used to return the endpoints registered by APIServer."""
         self.api_server.register_rest_endpoint(*options, **kwargs)
+
+    def configuration_endpoint(self):
+        """Return the configuration options used by Kytos.
+
+        Returns:
+            string: Json with current configurations used by kytos.
+        """
+        return json.dumps(self.options.__dict__)
 
     def restart(self, graceful=True):
         """Restart Kytos SDN Controller.
--- a/tests/test_core/test_api_server.py
+++ b/tests/test_core/test_api_server.py
@@ -1,0 +1,64 @@
+"""APIServer tests."""
+
+import logging
+import unittest
+
+from kytos.core.api_server import APIServer
+
+
+class TestAPIServer(unittest.TestCase):
+    """Test the class APIServer."""
+
+    def setUp(self):
+        """Instantiate a APIServer."""
+        self.api_server = APIServer('CustomName', False)
+
+    def test_register_rest_endpoint(self):
+        """Test whether register_rest_endpoint is registering an endpoint."""
+        self.api_server.register_rest_endpoint('/custom_method/',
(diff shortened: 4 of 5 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,8 @@ All notable changes to the kytos project will be documented in this file.
 ********************************
 Added
 =====
+- Added the rest endpoint '/kytos/configuration/' to display current
+  configuration loaded by Kytos.
```
cemsbr: Maybe it would be better a shorter description. Something like "- '/kytos/configuration/' endpoint to display kytos configuration.

kytos/core/api_server.py
```diff
@@ -83,30 +83,29 @@ def rest_endpoints(self):
         """Return string with routes registered by Api Server."""
         return [x.rule for x in self.app.url_map.iter_rules()]
 
-    def register_kytos_routes(self):
+    def register_api_server_routes(self):
         """Register initial routes from kytos using ApiServer.
 
-        Initial routes are: ['/kytos/status', '/kytos/shutdown']
+        Initial routes are: ['/kytos/status/', '/kytos/shutdown/', 'kytos/api']
```
cemsbr: Nice, trailing slash is the right way to do it. But where does kytos/api come from?

kytos/core/controller.py
```diff
@@ -89,14 +90,21 @@ def __init__(self, options=None):
         #: The key is the switch dpid, while the value is a Switch object.
         self.switches = {}  # dpid: Switch()
 
+        #: Attribute used to store the start_at time.
         self.started_at = None
```
cemsbr: We can be less verbose and more informative with, e.g.: ``#: datetime.datetime: Time when the controller finished starting.``

kytos/core/controller.py
```diff
@@ -89,14 +90,21 @@ def __init__(self, options=None):
...
+        #: Hash with all websockets used by Kytos.
         self.websockets = {}
 
+        #: Attribute used to write Kytos logs.
         self.log = None
 
+        #: API Server used to expose rest endpoints.
         self.api_server = APIServer(__name__, self.options.debug)
 
+        #: Create the rest endpoint /kytos/configuration
```
cemsbr: This method has its own docstring, so you shouldn't use attribute docstring.

kytos/core/controller.py
cemsbr: By the way, next time, try to specify the attribute type like my previous comment.

kytos/core/controller.py
```diff
@@ -213,10 +219,31 @@ def start_controller(self):
...
+
     def register_rest_endpoint(self, *options, **kwargs):
         """Method used to return the endpoints registered by APIServer."""
         self.api_server.register_rest_endpoint(*options, **kwargs)
 
+    def configuration_endpoint(self):
+        """Return the configuration options used by Kytos.
+
+        Returns:
+            string : Json with current configurations used by kytos.
```
cemsbr: Not sure whether it's part of the PEP, but let's keep using no spaces between ``string`` and ``:``.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+class TestAPIServer(unittest.TestCase):
+    """Test the class APIServer."""
+
+    def setUp(self):
+        """Instantiate a APIServer."""
+        self.api_server = APIServer('CustomName', False)
+
+    def test_register_rest_endpoint(self):
+        """Test whether register_rest_endpoint is registering a endpoint."""
```
cemsbr: a**n** endpoint.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+        expected_endpoint = '/kytos/custom_method/'
+        actual_endpoints = self.api_server.rest_endpoints
+        self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_register_api_server_routes(self):
+        """Test whether register_api_server_routes is working property."""
+        self.api_server.register_api_server_routes()
+
+        expecteds = ['/kytos/status/', '/kytos/shutdown/']
```
cemsbr: In English, adjectives have no plural form.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+        actual_endpoints = self.api_server.rest_endpoints
+
+        for expected_endpoint in expecteds:
+            self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_rest_endpoints(self):
+        """Test whether rest_endpoint return all endpoints registered."""
```
cemsbr: return**s** all **registered endpoints.** (adjective before noun)

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_register_rest_endpoint(self):
+        """Test whether register_rest_endpoint is registering a endpoint."""
+        def custom_method():
+            """Custom method used by APIServer."""
+            return 'Custom Method.'
+
+        self.api_server.register_rest_endpoint('/custom_method/',
+                                               custom_method,
+                                               methods=['GET'])
```
cemsbr: As you are not testing the endpoint results, can you simplify this test by using ``None`` as the endpoint method?

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+        expected_endpoint = '/kytos/custom_method/'
+        actual_endpoints = self.api_server.rest_endpoints
+        self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_register_api_server_routes(self):
+        """Test whether register_api_server_routes is working property."""
```
cemsbr: _working properly_ is too vague. After reading the test, I think _Server routes should include status and shutdown endpoints._ is a better description.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,66 @@
...
+        def custom_endpoint():
+            """Custom method used by APIServer."""
+            return "Custom Endpoint"
+
+        endpoints = ['/custom/', '/custom_2/', '/custom_3/']
+
+        for endpoint in endpoints:
+            self.api_server.register_rest_endpoint(endpoint,
+                                                   custom_endpoint,
+                                                   methods=['GET'])
```
cemsbr: Again, maybe you can just use ``None`` as the endpoint method. If not, create a new private static method instead of repeating ``def custom_endpoint...``.

tests/test_core/test_controller.py
```diff
@@ -0,0 +1,56 @@
...
+        actual_endpoints = self.controller.api_server.rest_endpoints
+        self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_register_kytos_endpoints(self):
+        """Verify all endpoints registered by Controller."""
+        expected_endpoints = ['/kytos/configuration/', '/kytos/shutdown/',
+                              '/kytos/status/', '/index.html', '/',
+                              '/static/<path:filename>']
+        actual_endpoints = self.controller.api_server.rest_endpoints
+        self.assertEqual(expected_endpoints, actual_endpoints)
```
cemsbr: ``self.assertListEqual`` gives better output on failure.

cemsbr (changes requested):

tests/test_core/test_controller.py
```diff
@@ -0,0 +1,56 @@
...
+        """Should register the endpoint '/kytos/configuration/'."""
+        expected_endpoint = '/kytos/configuration/'
+        actual_endpoints = self.controller.api_server.rest_endpoints
+        self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_register_kytos_endpoints(self):
+        """Verify all endpoints registered by Controller."""
+        expected_endpoints = ['/kytos/configuration/', '/kytos/shutdown/',
+                              '/kytos/status/', '/index.html', '/',
+                              '/static/<path:filename>']
```
diraol: `/`, `/static/<>` and `/index.html` are not endpoints from our REST API ....

diraol (changes requested):

tests/test_core/test_controller.py
macartur: So, today these endpoints is used to serve the web-ui.

diraol: @macartur I know  that our web-ui is served on this URLs, but they are not REST ENDPOINTS. They are just URLs. That's what I meant. @cemsbr, what I said is reasonable?

cemsbr: I'm not sure whether an URL is an endpoint, but we definitely don't have only REST endpoints in this list. Using "urls" instead of "endpoints" in this method wouldn't be wrong, because, certainly, a URL can be an endpoint.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,63 @@
...
+
+        expected_endpoint = '/kytos/custom_method/'
+        actual_endpoints = self.api_server.rest_endpoints
+        self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_register_api_server_routes(self):
+        """Server routes should include status and shutdown endpoints."""
+        self.api_server.register_api_server_routes()
+
+        expecteds = ['/kytos/status/', '/kytos/shutdown/']
```
cemsbr: My comment in the previous commit was not considered here.

tests/test_core/test_api_server.py
```diff
@@ -0,0 +1,63 @@
...
+        for expected_endpoint in expecteds:
+            self.assertIn(expected_endpoint, actual_endpoints)
+
+    def test_rest_endpoints(self):
+        """Test whether rest_endpoint returns all registered endpoints."""
+        endpoints = ['/custom/', '/custom_2/', '/custom_3/']
+
+        for endpoint in endpoints:
+            self.api_server.register_rest_endpoint(endpoint,
+                                                   self.__custom_endpoint,
```
cemsbr, line 43: My comment in the previous commit was not considered here.

cemsbr (changes requested):

cemsbr:
@macartur, not pushing a new commit makes the review process much longer. It is OK if you want to rebase, but do it after approval.

beraldoleal:
@macartur thanks... waiting for changes made by @diraol and @cemsbr . Also can you please rename to `/kytos/config/` ?

kytos/core/api_server.py
macartur: This was wrong, I removed this.

tests/test_core/test_api_server.py
macartur: Fixed.

tests/test_core/test_api_server.py
macartur: FIxed.

macartur:
I made all changes.

PR #450, opened 2017-06-13
Fix register_rest_endpoint method

### LINKED ISSUE #448: Registering endpoint REST with same URL but different method does not work
NApp [`of_flow_manager`](https://github.com/kytos/kytos-napps/blob/master/napps/kytos/of_flow_manager/main.py#L23) tries to register the same URL for GET, POST and DELETE with different function handlers, but it does not work because method [`register_rest_endpoint`](https://github.com/kytos/kytos/blob/master/kytos/core/api_server.py#L41) in `APIServer` class does not allow the register of the same URL more than once, even for different HTTP methods.

renanrodrigo:
Fix #448
Check URL and method before denying a new rest endpoint.
Log a warning message if an endpoint is not registered.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ All notable changes to the kytos project will be documented in this file.
 Added
 =====
 - '/kytos/configuration/' endpoint to display kytos configuration.
-- Add 'api_port' in kytos.conf, with default value equals 8181.
+- Add 'api_port' in kytos.conf. The default value is 8181.
 
 Changed
 =======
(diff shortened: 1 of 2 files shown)


diraol (approved):

diraol:
@renanrodrigo anything to add to the CHANGELOG?

renanrodrigo:
There's no entry for bug fixes yet, so I'm adding it. Thanks.

beraldoleal:
@renanrodrigo thanks... but can you rebase please.

PR #453, opened 2017-06-13
Improve kytos documentation

macartur:
Add blueprints tree
Fix sphinx warnigs: blueprints, administrator and kytos.core
Remove unused administrator/architechture.rst file
Remove nonlocal_uri warnings
Update CHANGELOG.rst

PR #444, opened 2017-06-05
Add dropdown button into navbar.html

macartur:

PR #455, opened 2017-06-14
Improve changelogs and update versions for release

diraol:
This PR also incorporates the PRs kytos/kytos#453 and kytos/kytos#444.
The first one improved kytos documentation and the second one add a dropdown
on kytos docs sphinx in order to allow multiple docs versions.


diraol:
I'm assuming that the tests for this PR are failing because I've updated the requirements.txt to contain `python-openflow>=2017.1b3` instead of `python-openflow>=2017.1b2`. As python-openflow>=2017.1b3 wasn't released on pypi, then the build will not work.

Issue #465, opened 2017-06-20
setup.py uses pip in a bad way


erickvermot:
Pip was meant to be called from the command line, and not from within a python session that plans to use the libraries it installs. It has no programmatic API. (you can see that in it's docs there is no reference to how to use it from a python session, and there is no "install()" method for example. calling pip.main(['install', ...]) is simply a bad workaround). See more about this discussion here: https://github.com/pypa/pip/issues/3121

Currently setup.py is calling pip by importing it and calling it's main() method as if it has been called from the command line, but since pip was not meant to be called in that way and expects the session to be closed afterwards, it does not update variables needed later (sys.modules cache for example), and may lead to import issues.

The correct way to use pip is from the command line, with its own python session.
That are a few fixes/alternatives to this:

1. - Use subprocess.check_call whenever we want to invoke pip.
     This is implemented here: https://github.com/erickvermot/kytos/tree/new_pip_call
     This fixes some of the potential problems, but is more of a workaround, and still leaves some potential issues, specially if the session that made the call tries to use/import the libraries installed.

2. - Use the method (1) and place all the methods that depend on potentially installed libs in other scripts, and call each of these scripts using its own subprocess.check_call.
     This is implemented here: https://github.com/erickvermot/kytos/tree/new_pip_call_separate_logics

3. - Make setup.py check_call pip to install the dependencies, and then check_call itself again with some parameter that prevents the pip call. (this method is used by some libs). I'll implement it if needed.

### To reproduce the error:
```
$ sudo docker run -it --privileged base/archlinux /bin/bash
[root@a211dfca7455 /]# pacman -Sy python-pip
[root@a211dfca7455 /]# pip install kytos

Collecting kytos
  Downloading kytos-2017.1b3.tar.gz (1.7MB)
    100% |################################| 1.7MB 288kB/s 
Installing collected packages: kytos
  Running setup.py install for kytos ... error
    Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-as8iljob/kytos/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-smxaedzp-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running bdist_egg
    running egg_info
    Installing dependencies...
    Collecting ipython
      Downloading ipython-6.1.0-py3-none-any.whl (741kB)
[...]

PR #469, opened 2017-06-21
call pip using subprocess to avoid sys.modules cache problems

erickvermot:
temporary solution to https://github.com/kytos/kytos/issues/465.


cemsbr:
Fixed tests in #470.

erickvermot:
#470 includes tests fixes that have no relation to this PR, and needed to be done independently of this PR. I don't think this should have been closed.

beraldoleal:
@cemsbr ? close or open ?

PR #470, opened 2017-06-21
Fix tests

cemsbr:
Add test fixes for #469.

**Edit**: only fix tests; removed fix for #469.

### DIFF
--- a/kytos/core/logs.py
+++ b/kytos/core/logs.py
@@ -112,8 +112,10 @@ class LogManager:
 
 
 # Add filter to all pre-existing handlers
+handler_filter = LogManager.filter_session_disconnected
 for root_handler in logging.root.handlers:
-    root_handler.addFilter(LogManager.filter_session_disconnected)
+    if handler_filter not in root_handler.filters:
+        root_handler.addFilter(handler_filter)
 
 
 class NAppLog:
--- a/tests/test_core/test_logs.py
+++ b/tests/test_core/test_logs.py
@@ -1,10 +1,12 @@
 """Test the logs module."""
+import imp
 import logging
 from copy import copy
 from inspect import FrameInfo
 from unittest import TestCase
 from unittest.mock import Mock, patch
 
+from kytos.core import logs
 from kytos.core.logs import LogManager, NAppLog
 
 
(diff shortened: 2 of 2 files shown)


diraol:
@cemsbr:
```
setup.py:17:4: W0611 Unused import pip [pylint]
kytos/core/logs.py:80:8: W0621 Redefining name 'handler' from outer scope (line 116) [pylint]
kytos/core/logs.py:85:25: W0621 Redefining name 'handler' from outer scope (line 116) [pylint]
```

erickvermot:
Sorry, but I don't understand...
What does these test fixes have to do with the Pip call?
Did it break because of #469, or was it broken before? I do not understand the relation to one another and why #469 was closed to be placed together with these other commits...

cemsbr:
PRs must pass in tests to be accepted. Another PR was created because you were not online to accept a PR to your branch. But don't worry, your commit is still there. Try to remember to run "tox" before submitting a PR, so you can detect failures before us.

The tests broke exactly in your commit. It seems that `pip.main` call changes `logging.root.handlers`, and now that it is gone, the tests broke. I made the failing tests more isolated from the environment to avoid future problems.

Thanks for the short and quick solution.

erickvermot:
As I see it, the tests had a bug that was exposed. You could/should open an issue about it, fix it in a first PR, and once that PR is accepted, tests would be working OK, and finally my PR #469 would pass the tests with no problem.
There is no reason to put it all together in a single PR, and it goes against what you guys have been preaching the last few months.

I'll try to be more clear on this one.

> PRs must pass in tests to be accepted.

agreed.

> Another PR was created because you were not online to accept a PR to your branch.

I don't think there was need for a PR to my branch since my branch was not the cause of the problem, and your fixes did not need my commit either.

> But don't worry, your commit is still there.

I'm not worried about the commit being there. I'm worried about the fact that putting it all together makes it look like the commit was the source of the test problems you solved, when it was not.

> Try to remember to run "tox" before submitting a PR, so you can detect failures before us.

I did run, but since my commit was not the cause of the problem, I was hoping you guys would fix the tests in another PR before accepting my PR...

> The tests broke exactly in your commit.

They did not break in my commit. They were broken before, and my commit only exposed the problem. It would have to be fixed eventually independently of my commit.

> It seems that pip.main call changes logging.root.handlers, and now that it is gone, the tests broke. I made the failing tests more isolated from the environment to avoid future problems.

You confirm what I said, the tests were not sufficiently isolated, and needed fix, regardless of my commit.

> Thanks for the short and quick solution.

You're welcome.

cemsbr:
You are correct. I just took the fastest route for this fix to be merged into master.

erickvermot:
the solution we discussed (https://github.com/kytos/kytos/issues/465#issuecomment-310296553) needs requires the fixes before it can pass the tests, so it'd be nice to have this fix in a separate branch/PR so I can use it in the other implementations.

diraol (approved):

cemsbr:
Erick is working on another approach that will be ready in the next minutes. This will only fix the tests so his PR will pass.

Issue #485, opened 2017-06-28
Admin Guide


cemsbr:

erickvermot:
for those wishing to follow and comment: https://github.com/erickvermot/kytos/tree/docs_updates

PR #492, opened 2017-06-29
Partial Docs updates

erickvermot:
partial update to https://github.com/kytos/kytos/issues/485


cemsbr (approved):
There are small issues that are easier to fix in a new commit by me than writing comments here.

beraldoleal:
I'm going to accept this but we need to discuss better the structure of this document.

PR #495, opened 2017-07-02
Fix docs: missing parameter description

macartur:


diraol:
@macartur Please, use a better PR title and description.

And do not forget to do the PR to both `master` and `2017.1rc` branches.

One more thing, why ignoring `D301`?

cemsbr (changes requested):
I wouldn't ignore D301, but use r""" where it happens. It is OK to rebase in this specific case. Please, write a comment like "Close #441" in a commit long description.

kytos/core/buffers.py
```diff
@@ -4,7 +4,7 @@
 
 from kytos.core.events import KytosEvent
 
-__all__ = ('KytosBuffers', )
+__all__ = ('KytosBuffers', 'KytosEventBuffer')
```
diraol: `KytosEventBuffer` was not supposed to be exposed.

kytos/core/buffers.py
```diff
@@ -29,6 +29,10 @@ def put(self, event):
 
         Reject new events is True when a kytos/core.shutdown message was
         received.
+
+        Args:
+            event (:class:`~kytos.core.buffers.KytosEventBuffer`):
```
diraol: event is an instance of `KytosEvent`, a class from `kytos.core.events` module.

kytos/core/buffers.py
```diff
@@ -40,7 +44,12 @@ def put(self, event):
             self._reject_new_events = True
 
     def get(self):
-        """Remove and return a event from top of queue."""
+        """Remove and return a event from top of queue.
+
+        Returns:
+            :class:`~kytos.core.buffers.KytosEventBuffer`:
```
diraol: This is not the correct class.
`kytos.core.buffers.KytosEventBuffer` != `kytos.core.events.KytosEvent`

kytos/core/buffers.py
```diff
@@ -29,6 +29,10 @@ def put(self, event):
 
         Reject new events is True when a kytos/core.shutdown message was
         received.
+
+        Args:
+            event (:class:`~kytos.core.buffers.KytosEventBuffer`):
+                KytosEventBuffer sent to queue.
```
diraol: Wrong here too.

kytos/core/buffers.py
```diff
@@ -83,18 +92,25 @@ class KytosBuffers(object):
...
+        sent to NApps.
         """
         self.raw = KytosEventBuffer('raw_event')
         self.msg_in = KytosEventBuffer('msg_in_event')
         self.msg_out = KytosEventBuffer('msg_out_event')
         self.app = KytosEventBuffer('app_event')
 
     def send_stop_signal(self):
-        """Method usd to send a kytos/core.shutdown event to all buffers."""
+        """Method to send a ``kytos/core.shutdown`` event to all buffers."""
```
diraol: Since we are rewriting, let's go with:
`Send a ``kytos/core.shutdown`` event to each buffer.`

kytos/core/switch.py
```diff
@@ -373,6 +391,12 @@ def last_flood(self, ethernet_frame):
         """Return the timestamp when the ethernet_frame was flooded.
 
         This method is usefull to check if a frame was flooded before or not.
+
+        Args:
+           ethernet_frame (:class:`pyof.foundation.basic_types.Ethernet`):
+               Ethernet instance to be verified.
+        Returns:
+           datetime.datetime.now: Last time when the ethernetframe was flooded.
```
diraol: `s/ethernetframe/ethernet_frame/`

kytos/core/tcp_server.py
```diff
@@ -30,10 +30,10 @@ def __init__(self, server_address, RequestHandlerClass, controller):
         Args:
             server_address (tuple): Address which the server is listening.
                 example: ('127.0.0.1', 80)
-            RequestHandlerClass (RequestHandlerClass):
+            RequestHandlerClass(socketserver.BaseRequestHandler):
                 Class that will be instantiated to handle each request.
-            controller (KytosController): The controller instance.
-
+            controller (:class:`~kytos.core.controller.Controller`):
+                The controller instance.
```
diraol: `s/The controller instance/An instance of Kytos Controller class/`

diraol (changes requested):

diraol (approved):
I think that this is the best solution for now, despite some disadvantages. As we intend to remove all python-openflow dependencies from kytos, I recommend this PR to be accepted to solve some documentation/linter warnings and errors.

PR #502, opened 2017-07-05
Fix docs: missing parameter description 

macartur:


diraol (approved):
I think that this is the best solution for now, despite some disadvantages. As we intend to remove all python-openflow dependencies from kytos, I recommend this PR to be accepted to solve some documentation/linter warnings and errors.

PR #527, opened 2017-07-11
Fix NApp unloading

cemsbr:
- Create specific NApp shutdown event "kytos/core.shutdown.username/napp"
- Call listener of the above when unloading a NApp


diraol:
@cemsbr why is there a commit from @macartur on your PR?

diraol (approved):

PR #533, opened 2017-07-18
Make KytosNApp().__event an instance attribute

### LINKED ISSUE #532: __event.set() causes re-enabled NApps's loops to fail.
When a NApp is disabled, a call to `self.__event.set()` is executed. If it is enabled again without restarting the controller, `self.__event.is_set()` keeps returning `True`, preventing the `execute` method to loop when intended.

A call to `self.__event.clear()` is needed prior to execution to fix this.

renanrodrigo:
Fix #532

### DIFF
--- a/kytos/core/napps/base.py
+++ b/kytos/core/napps/base.py
@@ -148,8 +148,6 @@ class NApp:
 class KytosNApp(Thread, metaclass=ABCMeta):
     """Base class for any KytosNApp to be developed."""
 
-    __event = Event()
-
     def __init__(self, controller, **kwargs):
         """Contructor of KytosNapps.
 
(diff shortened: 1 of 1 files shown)


diraol (commented):
@renanrodrigo I'm not absolutely sure that this is the best approach.

First, because I would expect that when a NApp is unloaded the NApp instance should be 'destroyed', then, when it is loaded again it should be a new instance. Therefore, it would behave as the first NApp load.

From the behavior you described, it looks like the NApp instance is being reused, which shouldn't be the case. So, I think we have do to a better investigation on why this is happening in order to really understand where the problem is. Specially because this can bring other undesired consequences.

Secondly, If we need to call `self.__event.clear()`, I would think on calling it on the `__init__()` method. It makes more sense to me to call it over there.

Btw, a fast look at the source coude made me think that this unexpected behavior may be happening because we are dealing with a Class attribute (`__event`) instead of an Instance attribute. If this is the reason, then we should instantiate `__event` on the `__init__()`.
I would do some tests to see if this won't bring problems related to our threading model while accessing the `__event` attribute.

Doing so, we probably won't need to call `__event.clear()`.

renanrodrigo:
Right, @diraol !
I'm testing thins and updating ASAP.

diraol (changes requested):
@renanrodrigo tested my suggestion and it seemed to work ok (add `self.__event = Event()` on the `__init__` instead of as class attribute. So, he will make the change proposed on my previous comment.

renanrodrigo:
As discussed above, I changed the commit to apply the most reasonable solution.

diraol (approved):

PR #534, opened 2017-07-18
Admin Guide Docs update 2

erickvermot:


erickvermot:
I'll close it because there is already another PR waiting that alters these files:
https://github.com/kytos/kytos/pull/530

Issue #536, opened 2017-07-18
Documentation Updates - Hacking/Dev guide


diraol:
Merge `hacking` (it is under kytos/documentation repo) with the `developer guide`.

erickvermot:
@diraol there is no ``hacking`` in documentation repo, do you mean hacking in the kytos/docs ?

erickvermot:
I'm unassigning myself because there seems to be nothing more useful from hacking.rst (it was written in the very early stages of dev), and the devguide structure is not decided yet.
Work can be continued once the structure is established.

PR #539, opened 2017-07-18
update docs/developer/hacking.rst

erickvermot:
https://github.com/kytos/kytos/issues/536
I took what I thought was important from docs/hacking.rst

### DIFF
--- a/docs/developer/hacking.rst
+++ b/docs/developer/hacking.rst
@@ -1,0 +1,51 @@
+#######
+Hacking
+#######
+
+.. note:: Before reading this file, please read the :doc:`kytos:contributing`
+    section that contains the main guidelines of the project.
+
+Development Environment Setup
+*****************************
+
+It is possible to install any of the *kytos* repos in develop mode, so that the
+installed files are kept in the original location, and only links are made to
+the system locations. This allows developers to quickly test changes made in
+the source code.
+
+To install a repo in develop mode, inside the repo source folder run:
+
+	.. code:: shell
+
+	    sudo pip3 install -e .
+
+
+TDD (Test Driven Development)
+*****************************
+
+We aim at 100% of test coverage. We are using
+Python `unittest <https://docs.python.org/3.5/library/unittest.html>`__ to
+write tests and
+`coverage.py <https://coverage.readthedocs.org/en/coverage-4.0.3/>`__ for
+coverage metrics. To install the coverage (python3 version), run:
+
+.. code:: shell
+
+    pip3 install coverage
+
+To run the tests, use the following command on the root folder of the project:
+
+.. code:: shell
+
+    python3 setup.py test
+
+To run check the code test coverage, first run:
+
+.. code:: shell
+
+    coverage run setup.py test
+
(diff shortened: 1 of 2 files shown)


docs/developer/hacking.rst
```diff
@@ -0,0 +1,51 @@
...
+It is possible to install any of the *kytos* repos in develop mode, so that the
+installed files are kept in the original location, and only links are made to
+the system locations. This allows developpers to quickly test changes made in
+source code.
+
+To install a repo in develop mode, inside the repo source folder run:
+
+	.. code:: shell
+
+	    sudo python setup.py develop
```
diraol: @cemsbr what about `pip install -e .`? Should we focus on installation process using `pip` instead of `setup`? As far as I remember this has something to do with wheels, doesn't it?

erickvermot: @diraol I have no idea if it has something to do with wheels... About the installation process, I believe it is not up to me to answer.

cemsbr: pip is now required, so we can use it instead. However, I think they are behaving similarly.

cemsbr: @erickvermot, you can find the installation process in the rest of the doc. Others have reviewed it.

erickvermot: @cemsbr, I reviewed it on quickstart for ex.
Still see no reason not to call setup. Result is the same, pip use setuptools.
On this same page we call setup to do the tests for ex.
I made the changes anyway.

cemsbr (approved):

docs/developer/hacking.rst
```diff
@@ -0,0 +1,51 @@
...
+
+.. note:: Before reading this file, please read the :doc:`kytos:contributing`
+    section that contains the main guidelines of the project.
+
+Development Environment Setup
+*****************************
+
+It is possible to install any of the *kytos* repos in develop mode, so that the
+installed files are kept in the original location, and only links are made to
+the system locations. This allows developpers to quickly test changes made in
```
diraol: @erickvermot Sorry, I didn't see this type on the first review.
`s/developpers/developers`.

docs/developer/hacking.rst
```diff
@@ -0,0 +1,51 @@
...
+source code.
```
diraol: Another one I missed on the review. `changes made in source code` -> `changes made in the source code`.

docs/developer/index.rst
```diff
@@ -18,6 +18,8 @@ the `tutorials <http://tutorials.kytos.io/>`__ made by Kytos Team.
 
 .. include:: api.rst
 
+.. include:: hacking.rst
+
```
diraol: I would also add `hacking` prior to the `api` and the `listened_events`. It makes more sense, doesn't it?

diraol (changes requested):

diraol (approved):

erickvermot:
As requested by @diraol and @cemsbr, I joined the review commits with the original one and force pushed.

Issue #538, opened 2017-07-18
DevGuide - Python Openflow Section


diraol:
On the `DevGuide` we need to have a "section" (page) `python-openflow`. On this "section" we need, at least, the following topics to be covered:
 - Basic lib organization;
 - How to use the python-openflow library;
   - pyof Objects;
   - packing;
   - unpacking;
 - Difference among "structs" and "messages";
 - How to code a new "struct" (on an existing pyof version);
 - How to code a new "message" (on an existing pyof version);
 - How to start a new "pyof version";
 - etc.

All of this topics must cover also how to implement the necessary tests to assure the qualitfy of the proposed code.

Suggestions to improve this section are welcome. What is missing here?

erickvermot:
@diraol, @cemsbr So python-openflow development section will be inside kytos docs devguide ?

erickvermot:
since python-openflow is a different repo, it seems to me it makes more sense to be inside the python-openflow repo...
Furthermore, much of the "How to code a new \*" is already inside the docstrings of the classes, like GenericStruct and GenericMessage

diraol:
@erickvermot for now we are concentrating all main documentation files from Kytos SDN Platform under 'kytos/docs'. Eventually, on the future, we may move this python-openflow documentation to pyof repository (under docs), 

So, doing it here inside a specific file will make it easy to move on the future.

Regarding the docstrings, yes, we do have some information about that topics over there, but here we can do a much better work, writing more texts, giving more examples and even using some images if needed. On the documentation we must master our explanations as much as we can. We want our (dev) users to reach this pages before they really need to go to the source code.

That does not mean that our source code docstrings doesn't need to be excelent, since they are used both on the specific "project API" documentation page and also on the command line.

erickvermot:
I have implemented 2 docs:

one with the current behavior of pyof:
  https://github.com/erickvermot/kytos/tree/docs_python_openflow

one with the behavior of pyof if some of the proposed mods, PRs, and open issues are accepted:
  https://github.com/erickvermot/kytos/compare/docs_python_openflow...erickvermot:docs_python_openflow_new_method

please take a look and let me know when I can open a PR.

erickvermot:
@cemsbr I did the changes you asked in your comments in the branch.
But are you sure it is easier or faster to make 17 comments about single character changes than change them while you review?

cemsbr:
[...]

PR #541, opened 2017-07-20
Docs python openflow

erickvermot:
https://github.com/kytos/kytos/issues/538

### DIFF
--- a/docs/developer/python-openflow.rst
+++ b/docs/developer/python-openflow.rst
@@ -20,6 +20,12 @@ Repository file tree
 Description
 +++++++++++
 
+Each openflow version is a subpackage inside the ``pyof`` package. It contains
+the version classes for messages and structures, which are organized according
+to the categories defined in the spec, like 'common', 'asynchronous', etc.
+Structures that are used by more than one category are usually placed in the
+'common' category.
+
 - docs/
     Sphinx documentation.
 
@@ -32,12 +38,12 @@ Description
     like the null terminated string ``foundation.basic_types.Char``.
 
 - pyof/v0x01/
-    Subpackage containing OpenFlow version 1.0 (0x01) classes for messages and
-    structs.
+    Subpackage containing OpenFlow version 1.0 (0x01) classes for messages,
+    structs and enumerations.
 
 - pyof/v0x04/
-    Subpackage containing OpenFlow version 1.3 (0x04) classes for messages and
-    structs.
+    Subpackage containing OpenFlow version 1.3 (0x04) classes for messages,
+    structs and enumerations.
 
 - raw/
     Raw OpenFlow message dumps to be used by the test suite.
@@ -114,6 +120,9 @@ To unpack a buffer as pyof object, simply initialize a new object and
 call its ``unpack()`` method, passing the buffer as the argument. The
 buffer will be unpacked in place setting the attributes of the object
 instance used to make the call.
+Exceptions to this case are ``GenericMessage``s derivatives, whose
+``unpack()`` method receives as argument a buffer containing only the
+message body.
 
 
 New definitions
@@ -127,16 +136,16 @@ Messages are GenericStruct derivatives who carry a header attribute containing
 an OpenFlow header struct.
 
 Unlike ``GenericStruct``s their unpack method accepts a buffer argument with
-its packed body (without the associated header).
+its binary body (without the associated header).
 
 How to code a new struct/message
 ++++++++++++++++++++++++++++++++
 To implement a new struct, you need to:
 - define a new class which derives from ``GenericStruct``;
-- define class attributes in pack order (with optional initial values) whose
-  type are pyof objects themselves.
+- define class attributes in following the order in which they must be packed
+  (with optional initial values) whose type are pyof objects themselves.
 
-Struct example:
+Struct example (``myexamplestruct.py``):
 
 .. code:: python
 
@@ -147,6 +156,10 @@ Struct example:
             my_first_attribute = UBInt8(255)
             my_second_attribute = UBInt16(0)
             my_third_attribute = UBInt8(255)
+
+            def __init__(self, my_third_attribute=None):
+                super().__init__()
+            	self.my_third_attribute = my_third_attribute
 
 which can be used like this:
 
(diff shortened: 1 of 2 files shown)


docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+    Main python-openflow module.
+
+- pyof/foundation/
+    Subpackage containing classes used/needed to implement a new pyof message
+    or structure, like ``foundation.base.GenericStruct`` and many basic types
+    like the null terminated string ``foundation.basic_types.Char``.
+
+- pyof/v0x01/
+    Subpackage containing OpenFlow version 1.0 (0x01) classes for messages and
+    structs.
```
diraol: `s/messages and structs/ messages, structs and enumerations/`
Is there any other "category" to be added here?

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+- pyof/v0x04/
+    Subpackage containing OpenFlow version 1.3 (0x04) classes for messages and
+    structs.
```
diraol: Same as the above comment.

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+  - pyof/
+    - foundation/
+    - v0x01/
+    - v0x04/
+  - raw/
+  - tests/
+
+Description
++++++++++++
+
```
diraol, line 22: I think that worth adding here a paragraph describing the source code structure regarding openflow versions. Like, saying that each version will be placed inside a specific directory and  that within each version messages and structures are divided according to the openflow spec "groups". Also, it worth mentioning that structures that are used by more than one "group" are usually placed at the "common" file or directory (depending if it is inside a "group" or used by other "groups" too). This may need a better explanation with some examples.

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
++++++++
+To pack a pyof object, simply call its ``pack()`` method, which returns a
+``bytes`` object of the binary representation.
+
+unpacking
++++++++++
+To unpack a buffer as pyof object, simply initialize a new object and
+call its ``unpack()`` method, passing the buffer as the argument. The
+buffer will be unpacked in place setting the attributes of the object
+instance used to make the call.
```
diraol, line 122: This does not work for "Messages", where the unpack receives the binary data without the message header.

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+``python-openflow`` provides some base classes and basic type classes meant to
+be used in a new message/struct class definition.
+
+Difference between "structs" and "messages"
++++++++++++++++++++++++++++++++++++++++++++
+Messages are GenericStruct derivatives who carry a header attribute containing
+an OpenFlow header struct.
+
+Unlike ``GenericStruct``s their unpack method accepts a buffer argument with
+its packed body (without the associated header).
```
diraol: `s/packed body/binary body/`

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+How to code a new struct/message
+++++++++++++++++++++++++++++++++
+To implement a new struct, you need to:
+- define a new class which derives from ``GenericStruct``;
+- define class attributes in pack order (with optional initial values) whose
```
diraol: `s/pack order/following the order in which they must be packed/`

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+.. code:: python
+
+    >>> from pyof.foundation.base import GenericStruct
+    >>> from pyof.foundation.basic_types import UBInt8, UBInt16
+
+    >>> class MyNewStruct(GenericStruct):
+            my_first_attribute = UBInt8(255)
+            my_second_attribute = UBInt16(0)
+            my_third_attribute = UBInt8(255)
```
diraol, line 158: Add also the `__init__` class passing the `__init__` arguments to show, on the example, that the instance can be created with the desired values on the object instantiation.

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+.. code:: python
+
+    >>> from pyof.foundation.base import GenericMessage
+    >>> from pyof.foundation.basic_types import UBInt8, UBInt16
+    >>> from pyof.v0x04.common.header import Header
+
+    >>> class MyNewMessage(GenericMessage):
+            header = Header(message_type=255)
+            my_struct_field = MyNewStruct()
```
diraol: `MyNewStruct` was not imported. neither defined together with `MyNewMessage`.

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+How to start a new "pyof version"
++++++++++++++++++++++++++++++++++
+
+- Create a new package under ``pyof`` named after the version you are
+  interested in implementing, like ``v0xff`` for example.
+- Create new subpackages for your message categories.
+- Implement your new message as described before.
+
+your file tree should look like this::
```
diraol: `s/your/Your/`

docs/developer/python-openflow.rst
```diff
@@ -0,0 +1,206 @@
...
+
+  - pyof/
+    + foundation/
+    - v0xff/
+      - my_message_category/
+        __init__.py
+        mynewmessage.py
+      __init__.py
```
diraol: add this `__init__.py` file before `my_message_category` to improve readability.

Also, if you are doing a "listing", it would be nice to put list markers on all lines.
The same may apply for the first listing from this file.

diraol (changes requested):
Besides the specific comments, I think that we may need to think about organizing this file for those who are going to just **use** the library (while developing a NApp) and for those who wants to write code for the lib.

The first group (only users) does not need to really understand  the difference between GenericStruct, GenericMessage and GenericType classes. [They need to know how to use their child classes]

Does this make sense?
@cemsbr what do you think about it?

Ow, and the PR is not passing on the tests.

docs/developer/python-openflow.rst
erickvermot: It is defined in the previous example, and it seems very  clear to me. I was avoiding being redundant.

docs/developer/python-openflow.rst
erickvermot: It is not a simple "listing", but a file tree. \- and \+ represent collapsed and expanded folders respectively. That's also why I list folders first, like it usually happens.

PR #544, opened 2017-07-21
Fix regex to download napp based on version.

macartur:


cemsbr (approved):
Ok to use a dash because this is not a valid character for both username and NApp name.

diraol (approved):

beraldoleal (changes requested):
I'm not sure if this will affect this PR but talking with @diraol and @cemsbr I would like to keep ":" as separator for versions. (Like docker).

Of course we can still use - to save our files.

But please keep in mind that our uri is with ":".

macartur:
I will close this PR because this change is unnecessary if we use ":" as a separator for versions.

Issue #521, opened 2017-07-07
Wildcards in Flow class


ajoaoff:
In our application, it will be necessary to match a packet against the flow in the control plane, for Interdomain tracepath. To do that, wildcards are necessary, but they are not available in [Flow](https://github.com/kytos/kytos/blob/master/kytos/core/flow.py#L11) class.

diraol:
@ajoaoff looking again to this issue, and also to #517 and #518, one thing has got my attention.

You are using the Flow class defined inside kytos core. But, as you are developing a NApp, it is better if you use the python-openflow library to build and 'use' OpenFlow strucutres.
As an example, you can use the classes [`FlowMod`](https://docs.kytos.io/python-openflow/pyof.v0x01.controller2switch/#pyof.v0x01.controller2switch.flow_mod.FlowMod) and [`Match`](https://docs.kytos.io/python-openflow/pyof.v0x01.common/#pyof.v0x01.common.flow_match.Match), which were implemented following the Openflow Specification.

Take a look at them and see if they fit your needs. =)

#edited

ajoaoff:
@diraol maybe not the best option. 
For example, we need to use the flows currently installed at the switches. The NApp kytos/core already uses FlowStats to retrieve the flows, but stores than using the Flow class. If I directly use python-openflow library, I would have to retrieve the flows again, and I don't know if it is a good idea.

cemsbr:
@macartur, could you please add the wildcard attribute as you did in the issues referenced by Diego? Because wildcards are automatically managed by pyof, you only have to add the wildcard when converting **to** JSON.

diraol:
Fixed on #511 .

@ajoaoff could you please test it to see if it is working ok?
Thanks!

cemsbr:
Without comments, I'm assuming it's solved. If not, please, reopen.

PR #551, opened 2017-08-03
Add wildcards attribute to Flow class

macartur:


kytos/core/flow.py
```diff
@@ -225,9 +230,9 @@ def as_flow_mod(self, flow_type=FlowModCommand.OFPFC_ADD):
                 setattr(flow_mod, attribute_name, value)
 
         flow_mod.match = Match()
-        match_attributes = ['in_port', 'dl_src', 'dl_dst', 'dl_vlan',
-                            'dl_vlan_pcp', 'dl_type', 'nw_src', 'nw_dst',
-                            'tp_dst']
+        match_attributes = ['wildcards', 'in_port', 'dl_src', 'dl_dst',
```
cemsbr: Because wildcards are automatically managed by pyof, we don't need to have ``wildcards`` here. Please, remove it to minimize the changes.

cemsbr (changes requested):
Please, also add a reference to issue #521 in commit message.

diraol (commented):
I don't have any other change to request for this PR. After its commits are squashed I think it can be merged onto master branch.

PR #552, opened 2017-08-08
Regular user can write in root's NApp folder

### LINKED ISSUE #543: Unable to run kytos as regular user even after fresh install
Since 07Jul, it is no longer possible to run kytos as a regular user.
It is now trying to create napps dir inside ``/var/lib`` during runtime, which is not possible by regular users.
https://github.com/kytos/kytos/commit/a8153507805444b78781ce090fcd14b37167bb55#diff-8bc31c20cd5303df1ce2ccc1de508fc0R32

Before that it was possible to run kytos normally as a regular user.
There was the problem of ``/var/run/kytos`` being deleted after reboot in some systems using recent versions of systemd, but everything worked ok with a fresh install. (see https://github.com/kytos/kytos/issues/381)

workaround is the same as for the other issue
``sudo mkdir -m 777 /var/lib/kytos``

cemsbr:
Temporary fix while we don't have a Kytos system group.

### DIFF
--- a/kytos/core/napps/napp_dir_listener.py
+++ b/kytos/core/napps/napp_dir_listener.py
@@ -29,7 +29,8 @@ class NAppDirListener(RegexMatchingEventHandler):
         super().__init__()
         self._controller = controller
         self.napps_path = self._controller.options.napps
-        Path(self.napps_path).mkdir(mode=0o755, parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


beraldoleal:
@cemsbr Thanks. I'm accepting this.. but this is just a workaround. In the future let's make sure that kytos-utils uses kytos controller API / RPC to ask to install NApps.

PR #560, opened 2017-09-05
Add @rest decorator for REST endpoints

cemsbr:
The code, including the decorator, is in APIServer. It mimics Flask
as much as possible by forwarding parameters. This keeps the code small
and the behavior consistent with Flask. Summary:
1. The decorator stores @rest params in function.route_params (same
   strategy as @listen_to);
2. APIServer.start_rest_api(napp) use the bound method and route_params
   exactly like Flask's @app.route does;
3. If a URL and an HTTP method combination is used twice or more, the
   user is warned in the logs and the behavior is the same as Flask's:
   the last declaration prevails.

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -148,8 +148,13 @@ class APIServer:
         APIServer and NApp instances.
         """
         def store_route_params(function):
-            """Store ``Flask`` ``@route`` parameters in a method attribute."""
-            function.route_params = rule, options
+            """Store ``Flask`` ``@route`` parameters in a method attribute.
+
+            There can be many @route decorators in a single function.
+            """
+            if not hasattr(function, 'route_params'):
+                function.route_params = []
+            function.route_params.append((rule, options))
             return function
         return store_route_params
 
@@ -159,8 +164,9 @@ class APIServer:
         URLs will be prefixed with ``/api/{username}/{napp_name}/``.
         """
         for method in self._get_decorated_methods(napp):
-            rule, options = method.route_params
-            self._start_endpoint(rule, method, **options)
+            for rule, options in method.route_params:
+                absolute_rule = self._get_absolute_rule(rule, napp)
+                self._start_endpoint(absolute_rule, method, **options)
 
     @staticmethod
     def _get_decorated_methods(napp):
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -66,21 +66,7 @@ class APIServer:
                 e.g: [``'GET'``, ``'PUT'``, ``'POST'``, ``'DELETE'``,
                 ``'PATCH'``]
         """
-        new_endpoint_url = "/kytos{}".format(url)
-
-        for endpoint in self.app.url_map.iter_rules():
-            if endpoint.rule == new_endpoint_url:
-                for method in methods:
-                    if method in endpoint.methods:
-                        message = ("Method '{}' already registered for " +
-                                   "URL '{}'").format(method, new_endpoint_url)
-                        self.log.warning(message)
-                        self.log.warning("WARNING: Overlapping endpoint was " +
-                                         "NOT registered.")
-                        return
-
-        self.app.add_url_rule(new_endpoint_url, function.__name__, function,
-                              methods=methods)
+        self._start_endpoint(url, function, methods=methods)
 
     def register_web_ui(self):
         """Method used to register routes to the admin-ui homepage."""
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -126,7 +126,7 @@ class APIServer:
     # BEGIN decorator methods
 
     @staticmethod
-    def decorator(rule, **options):
+    def decorate_as_endpoint(rule, **options):
         """Decorator for REST endpoints using Flask.
 
         Example for URL ``/api/myusername/mynapp/sayhello/World``:
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -8,9 +8,15 @@ from urllib.request import urlopen
 from flask import Flask, request, send_from_directory
 from flask_socketio import SocketIO, join_room, leave_room
 
+#: list: Default Flask HTTP methods used when no mehtods are specified by the
+# user. Used to warn multiple declarations of the same URL and HTTP method.
+_DEFAULT_METHODS = ('GET',)
+
 
 class APIServer:
     """Api server used to provide Kytos Controller routes."""
+
+    _NAPP_PREFIX = "/api/{napp.username}/{napp.name}/"
 
     def __init__(self, app_name, listen='0.0.0.0', port=8181):
         """Constructor of APIServer.
@@ -24,7 +30,7 @@ class APIServer:
         """
         dirname = os.path.dirname(os.path.abspath(__file__))
         self.flask_dir = os.path.join(dirname, '../web-ui')
-        self.log = logging.getLogger('werkzeug')
+        self.log = logging.getLogger('api_server')
 
         self.listen = listen
         self.port = port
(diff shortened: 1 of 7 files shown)


cemsbr:
This is to test Scrutinizer that wasn't triggered in the previous PR. I'm still working on it. These commits are finished and I'm refactoring the old way of starting an API.

kytos/core/api_server.py
```diff
@@ -130,3 +134,65 @@ def shutdown_api(self):
...
+
+    def _start_endpoint(self, rule, function, **options):
+        """Start ``function``'s endpoint.
+
+        1. Add '/kytos' prefix to the URL;
+        2. Warn if the same URL and method were already started.
+        2. Forward parameters to ``Flask.add_url_rule`` mimicking
+           Flask ``@route`` decorator.
+        """
+        rule = f'/kytos{rule}'
```
beraldoleal: Let's try to add `api/<author>/<napp_name>/` as prefix.

kytos/core/controller.py
```diff
@@ -582,6 +582,7 @@ def load_napp(self, username, napp_name):
             # This start method is inherited from the Threading class.
             # It is not directly defined/declared on the KytosNApp class.
             napp.start()
+            self.api_server.start_rest_api(napp)
```
beraldoleal: `s/start_rest_api/add_endpoints_from_decorators/g`

kytos/core/api_server.py
```diff
@@ -130,3 +134,65 @@ def shutdown_api(self):
...
+        object is passed as a parameter of :meth:`start_rest_api`.
+        """
+        def store_route_params(function):
+            """Store ``Flask`` ``@route`` parameters in a method attribute."""
+            function.route_params = rule, options
+            return function
+        return store_route_params
+
+    def start_rest_api(self, napp):
+        """Start ``napp`` REST endpoints."""
```
beraldoleal: Let's try to make APIServer NAPP agnostic/independent. `s/napp/instance/g`  for instance.

For the APIServer perspective doesn't matter if is a napp or other object, right ?

kytos/core/api_server.py
```diff
@@ -130,3 +134,65 @@ def shutdown_api(self):
...
+
+        1. Add '/kytos' prefix to the URL;
+        2. Warn if the same URL and method were already started.
+        2. Forward parameters to ``Flask.add_url_rule`` mimicking
+           Flask ``@route`` decorator.
+        """
+        rule = f'/kytos{rule}'
+        endpoint = options.pop('endpoint', None)
+        self._warn_multiple_api_defs(rule, options.get('methods'))
+        self.app.add_url_rule(rule, endpoint, function, **options)
```
beraldoleal: Since that my comments above asks for hardcode <author>/<napp_name> as prefix. If an user request two endpoints with same string, is not our problem.

Let's use the same flask behavior: let's not warn for now.. in order to keep this code simpler.

Because of that, you can remove the method `_warn_multiple_api_defs`

beraldoleal (changes requested):
Great work.. let's make these small changes and I will accept.

Please add few lines to changelog too.

kytos/core/api_server.py
```diff
@@ -130,3 +136,75 @@ def shutdown_api(self):
...
+
+    # BEGIN decorator methods
+
+    @staticmethod
+    def decorate_as_endpoint(rule, **options):
+        """Decorator for REST endpoints using Flask.
+
+        Parameter documentation is the same as Flask's @app.route. As we don't
+        have the NApp instance now, we store the parameters in a method
+        attribute in order to add the route later, after APIServer is the NApp
```
diraol: @cemsbr this sentence does not make sense:

> ...attribute in order to add the route later,  after APIServer is the NApp object is passed as a parameter of :meth:`start_rest_api`.

kytos/core/api_server.py
```diff
@@ -130,3 +136,75 @@ def shutdown_api(self):
     def web_ui(self):
         """Method used to serve the index.html page for the admin-ui."""
         return send_from_directory(self.flask_dir, 'index.html')
+
+    # BEGIN decorator methods
+
+    @staticmethod
+    def decorate_as_endpoint(rule, **options):
```
diraol, line 129: @cemsbr I didn't understand how this method works. Can you add an example on how to use it on the docstring, please?

kytos/core/api_server.py
```diff
@@ -130,3 +136,75 @@ def shutdown_api(self):
...
+        """Decorator for REST endpoints using Flask.
+
+        Parameter documentation is the same as Flask's @app.route. As we don't
+        have the NApp instance now, we store the parameters in a method
+        attribute in order to add the route later, after APIServer is the NApp
+        object is passed as a parameter of :meth:`start_rest_api`.
+        """
+        def store_route_params(function):
```
diraol: It wouldn't be better to used `method` instead of `function` here?

kytos/core/api_server.py
```diff
@@ -130,3 +136,75 @@ def shutdown_api(self):
...
+            """Store ``Flask`` ``@route`` parameters in a method attribute."""
+            function.route_params = rule, options
```
diraol: By doing this attribution, it is not possible to have two different endpoints on the same `function`/`method` right? I mean, we can not use our decorator twice on the same `method` (with flask `route` decorator this is possible). It wouldn't be better to have `route_params` as a list of tuples and then just append a new `(rule, options)` to it?

kytos/core/api_server.py
diraol: @beraldoleal I didn't understood you comment. What kind of object would we receive here to register an endpoint?
As far as I understood this method was conceived to be used on the NApps loading process, and, despite some specific corner cases from the core itself, there is no other kind of `object` that will register endpoints besides the NApps right?

Ow, and BTW, the decorator will not be used on the "core corner cases".

kytos/core/api_server.py
```diff
@@ -130,3 +136,75 @@ def shutdown_api(self):
...
+
+        # Look for the same rule and method
+        for rule_obj in self.app.url_map.iter_rules():
+            if rule_obj.rule == rule:
+                for started_method in rule_obj.methods:
+                    if started_method in methods:
+                        msg = f'API: more than one {started_method} {rule}'
+                        self.log.warning(msg)
+
+    # END decorator methods
```
diraol: Again this comment? There is another one on line 180.

kytos/core/api_server.py
diraol: This is strange...

kytos/core/api_server.py
diraol: @beraldoleal I think it is nice from our side to warn the user. And it the work is already done anyway.

Considering the case of a NApp with only one class (`Main`), the probability of the user overriding and endpoint is minimal, but considering a more complex NApp with more modules (such as of_stats), it is a little bit more possible to override something.

kytos/core/controller.py
diraol: I would rename it to `load_napp_endpoints` or `register_napp_endpoints`.

tests/test_core/test_api_server.py
```diff
@@ -53,3 +55,85 @@ def test_rest_endpoints(self):
...
+
+            @rest(rule, **options)
+            def my_endpoint(self):
+                """API method."""
+                pass
+
+        napp = MyNApp()
+        server = cls._mock_api_server(napp)
+        server.app.add_url_rule.assert_called_once_with(
+            '/kytos' + rule, None, napp.my_endpoint, **options)
```
diraol: Aren't we changing the endpoint to `/api/<napp.user>/<napp.name>/<rule>`?

diraol (changes requested):

cemsbr:
@beraldoleal, @diraol, I made the requested changes.

diraol (approved):

beraldoleal:
@cemsbr changelog ?

PR #596, opened 2017-09-29
Add method to remove rest endpoint

### LINKED ISSUE #508: Remove NApp REST Endpoints on NApp unload
While unloading a NApp, we already remove it's (event) listeners. We should also remove the REST Endpoints that were registered by the NApp when it was loaded.

macartur:
This commit will allow remove napps endpoints when a napp is disabled.

Fix #508

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -212,11 +212,16 @@ class APIServer:
             napp (Napp): Napp instance to look for rest-decorated methods.
         """
         prefix = self._NAPP_PREFIX.format(napp=napp)
+
+        indexes = []
         for index, rule in enumerate(self.app.url_map.iter_rules()):
             if rule.rule.startswith(prefix):
                 self.app.view_functions.pop(rule.endpoint)
-                # pylint: disable=protected-access
-                self.app.url_map._rules.pop(index)
-                # pylint: enable=protected-access
+                indexes.append(index)
                 self.log.info('Stopped %s - %s', rule, ','.join(rule.methods))
+
+        for index in reversed(indexes):
+            # pylint: disable=protected-access
+            self.app.url_map._rules.pop(index)
+            # pylint: enable=protected-access
         self.log.info(f'The Rest endpoints from %s were disabled.', prefix)
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -211,13 +211,13 @@ class APIServer:
         Args:
             napp (Napp): Napp instance to look for rest-decorated methods.
         """
-        napp_id = f'{napp.username}/{napp.name}'
+        prefix = self._NAPP_PREFIX.format(napp=napp)
         for index, rule in enumerate(self.app.url_map.iter_rules()):
-            if rule.rule.startswith(f'/api/{napp_id}'):
+            if rule.rule.startswith(prefix):
                 self.app.view_functions.pop(rule.endpoint)
                 # pylint: disable=protected-access
                 self.app.url_map._rules.pop(index)
                 # pylint: enable=protected-access
                 self.log.info('Stopping endpoint %s - %s', rule,
                               ','.join(rule.methods))
-        self.log.info(f'The Rest endpoints from %s were disabled.', napp_id)
+        self.log.info(f'The Rest endpoints from %s were disabled.', prefix)
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -163,6 +163,9 @@ class APIServer:
         """Add all NApp REST endpoints with @rest decorator.
 
         URLs will be prefixed with ``/api/{username}/{napp_name}/``.
+
+        Args:
+            napp (Napp): Napp instance to register new endpoints.
         """
         for function in self._get_decorated_functions(napp):
             for rule, options in function.route_params:
@@ -201,3 +204,20 @@ class APIServer:
         self.app.add_url_rule(rule, endpoint, function, **options)
         self.log.info('Started %s - %s', rule,
                       ', '.join(options.get('methods', self.DEFAULT_METHODS)))
+
+    def remove_napp_endpoints(self, napp):
+        """Remove all decorated endpoints.
+
+        Args:
+            napp (Napp): Napp instance to look for rest-decorated methods.
+        """
+        napp_id = f'{napp.author}/{napp.name}'
+        for index, rule in enumerate(self.app.url_map.iter_rules()):
+            if rule.rule.startswith(f'/api/{napp_id}'):
+                del self.app.view_functions[rule.endpoint]
+                # pylint: disable=protected-access
+                self.app.url_map._rules.pop(index)
+                # pylint: enable=protected-access
+                self.log.info('Stopping endpoint %s - %s', rule,
+                              ','.join(rule.methods))
+        self.log.info(f'The Rest endpoints from %s were disabled.', napp_id)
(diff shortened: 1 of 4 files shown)


kytos/core/api_server.py
```diff
@@ -59,6 +59,24 @@ def register_rest_endpoint(self, url, function, methods):
...
+
+        """
+        for index, rule in enumerate(self.app.url_map.iter_rules()):
+            if napp_id in rule.rule:
+                del self.app.view_functions[rule.endpoint]
+                # pylint: disable=protected-access
+                self.app.url_map._rules.pop(index)
+                # pylint: enable=protected-access
+
+        self.log.info(f'The Rest endpoints form {napp_id} was disabled.')
```
diraol: `s/form/from/`
`s/was/were/`

kytos/core/api_server.py
```diff
@@ -59,6 +59,24 @@ def register_rest_endpoint(self, url, function, methods):
             url = url[1:]
         self._start_endpoint(f'/kytos/{url}', function, methods=methods)
 
+    def remove_rest_endpoints(self, napp_id):
```
diraol: `s/remove_rest_endpoint/remove_napp_rest_endpoints`
(This method is designed to remove endpoints related to `napp_id`)

kytos/core/api_server.py
```diff
@@ -59,6 +59,24 @@ def register_rest_endpoint(self, url, function, methods):
...
+        """Remove a rest endpoint from flask.
+
+        This method will remove all endpoints that contain the napp_id.
+
+        Args:
+            napp_id: String using the format '<username>/<napp_name>'
+
+        """
+        for index, rule in enumerate(self.app.url_map.iter_rules()):
+            if napp_id in rule.rule:
```
diraol: I  think it would be better to check if `rule.rule` "starts" with `/api/<napp_id>` instead of just 'containing' it.

Because if `napp_id = "kytos/test"` and `rule.rule = "/api/diraol/test42/kytos/test"` would create a false positive.

diraol (changes requested):

kytos/core/api_server.py
cemsbr: @diraol, I think it could be even better if it behaves exactly like the `register_napp_endpoints`, so we are sure that all and only the endpoints started by the decorator will be removed.

kytos/core/api_server.py
cemsbr: Please, use the same message style as when starting the endpoint.

kytos/core/api_server.py
```diff
@@ -59,6 +59,24 @@ def register_rest_endpoint(self, url, function, methods):
...
 
+    def remove_rest_endpoints(self, napp_id):
+        """Remove a rest endpoint from flask.
+
+        This method will remove all endpoints that contain the napp_id.
+
+        Args:
+            napp_id: String using the format '<username>/<napp_name>'
+
+        """
```
cemsbr:
```
Remove all decorated endpoints.

Args:
    napp (NApp): NApp instance to look for rest-decorated methods.
```
By receiving a NApp instance, you can code like `register_napp_endpoints` to get the same endpoints.

kytos/core/controller.py
```diff
@@ -632,6 +632,9 @@ def unload_napp(self, username, napp_name):
             # Call listener before removing it from events_listeners
             napp_shutdown_fn(event)
 
+            # Remove rest endpoints from that napp
+            self.api_server.remove_rest_endpoints(napp_id)
```
cemsbr: Maybe using the napp instance will make things easier.

cemsbr (changes requested):

diraol (changes requested):
@macartur can you please add some tests here?
;)

diraol (approved):

diraol (changes requested):
Ow, and please add some message to the CHANGELOG!

diraol (approved):

kytos/core/api_server.py
```diff
@@ -198,3 +201,20 @@ def _start_endpoint(self, rule, function, **options):
...
+
+    def remove_napp_endpoints(self, napp):
+        """Remove all decorated endpoints.
+
+        Args:
+            napp (Napp): Napp instance to look for rest-decorated methods.
+        """
+        napp_id = f'{napp.username}/{napp.name}'
+        for index, rule in enumerate(self.app.url_map.iter_rules()):
+            if rule.rule.startswith(f'/api/{napp_id}'):
```
cemsbr: Reuse code wherever you can:
```python
prefix = cls._NAPP_PREFIX.format(napp=napp)
# ...
    if rule.rule.startswith(prefix):
```

tests/test_core/test_api_server.py
```diff
@@ -64,6 +64,37 @@ def my_endpoint(self):
...
+        server = cls._mock_api_server(napp)
+
+        rule = Mock()
+        rule.methods = ['GET', 'POST']
+        rule.rule.startswith.return_value = True
+        endpoint = 'username/napp_name'
+        rule.endpoint = endpoint
+
+        server.app.url_map.iter_rules.return_value = [rule]
+        server.app.view_functions.pop.return_value = True
```
cemsbr: `pop` returns the element itself (`rule`).

tests/test_core/test_api_server.py
```diff
@@ -64,6 +64,37 @@ def my_endpoint(self):
...
+        # pylint: disable=protected-access
+        server.app.url_map._rules.pop.return_value = True
```
cemsbr: The same here.

kytos/core/api_server.py
```diff
@@ -198,3 +201,20 @@ def _start_endpoint(self, rule, function, **options):
...
+                self.app.view_functions.pop(rule.endpoint)
+                # pylint: disable=protected-access
+                self.app.url_map._rules.pop(index)
```
cemsbr: Are you sure there won't be a problem if we pop while iterating? Maybe you could write a test for this situation.

kytos/core/api_server.py
```diff
@@ -198,3 +201,20 @@ def _start_endpoint(self, rule, function, **options):
...
+                # pylint: enable=protected-access
+                self.log.info('Stopping endpoint %s - %s', rule,
+                              ','.join(rule.methods))
+        self.log.info(f'The Rest endpoints from %s were disabled.', napp_id)
```
cemsbr: I believe this is unnecessary. You can remove this line.

cemsbr (changes requested):

kytos/core/api_server.py
```diff
@@ -198,3 +201,20 @@ def _start_endpoint(self, rule, function, **options):
...
+            napp (Napp): Napp instance to look for rest-decorated methods.
+        """
+        napp_id = f'{napp.username}/{napp.name}'
+        for index, rule in enumerate(self.app.url_map.iter_rules()):
+            if rule.rule.startswith(f'/api/{napp_id}'):
+                self.app.view_functions.pop(rule.endpoint)
+                # pylint: disable=protected-access
+                self.app.url_map._rules.pop(index)
+                # pylint: enable=protected-access
+                self.log.info('Stopping endpoint %s - %s', rule,
```
cemsbr: Stopped %s - %s

cemsbr (changes requested):

cemsbr (approved):

Issue #201, opened 2017-01-08
Can't change listening port


cemsbr:
This is a user feedback. He said that the port is always 6633, even if you change it.

cemsbr:
Closing as a consequence of #216.

PR #597, opened 2017-10-01
Fix the tcp_server to listen to any OpenFlow port

viniarck:
Hi guys,

I noticed that any TCP ports other than 6633 wasn't working properly because there is a hardcoded value on `kytos/core/tcp_server.py` that always assumed that the connection would be on 6633 (issue #201 apparently was supposed to fix this). So, I also added the port number 6653 to the `known_ports` OpenFlow dict. Still, other ports than these two wouldn't work. I believe you guys coded this way to add support for new southbound propocols, however, since at the moment only OpenFlow is supported, I changed the fallback else clause to assume that any other port would be OpenFlow. As a result, the user can use any TCP port he wants. 

Without this fix, as you can see bellow, the OpenFlow connection handshake keeps flapping:

- listening on port `6653`:

```
OF Address....: tcp://0.0.0.0:6653
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-01 12:37:35,228 - INFO [root] (kyslacker) Running NApp: <Main(kyslacker, started 140284126013184)>
2017-10-01 12:37:35,230 - INFO [root] (of_lldp) Running NApp: <Main(of_lldp, started 140284117620480)>
2017-10-01 12:37:35,230 - INFO [root] (of_topology) Running NApp: <Main(of_topology, started 140283563996928)>
2017-10-01 12:37:35,235 - INFO [root] (kytos/of_core) Running NApp: <Main(kytos/of_core, started 140283555604224)>
2017-10-01 12:37:35,751 - INFO [kytos.core.tcp_server] (Thread-12) New connection from 127.0.0.1:39972
2017-10-01 12:37:35,758 - INFO [kytos.core.tcp_server] (Thread-13) New connection from 127.0.0.1:39974
2017-10-01 12:37:35,763 - INFO [kytos.core.tcp_server] (Thread-14) New connection from 127.0.0.1:39976
2017-10-01 12:37:35,764 - INFO [kytos.core.tcp_server] (Thread-15) New connection from 127.0.0.1:39978
2017-10-01 12:37:36,751 - INFO [kytos.core.tcp_server] (Thread-12) Connection lost with Client127.0.0.1:39972. Reason: Request closed by client.
2017-10-01 12:37:36,752 - INFO [kytos.core.tcp_server] (Thread-15) Connection lost with Client127.0.0.1:39978. Reason: Request closed by client.
2017-10-01 12:37:36,753 - INFO [kytos.core.tcp_server] (Thread-14) Connection lost with Client127.0.0.1:39976. Reason: Request closed by client.
2017-10-01 12:37:36,754 - INFO [kytos.core.tcp_server] (Thread-13) Connection lost with Client127.0.0.1:39974. Reason: Request closed by client.
2017-10-01 12:37:37,752 - INFO [kytos.core.tcp_server] (Thread-16) New connection from 127.0.0.1:39980
2017-10-01 12:37:37,754 - INFO [kytos.core.tcp_server] (Thread-17) New connection from 127.0.0.1:39982
2017-10-01 12:37:37,755 - INFO [kytos.core.tcp_server] (Thread-18) New connection from 127.0.0.1:39984
2017-10-01 12:37:37,757 - INFO [kytos.core.tcp_server] (Thread-19) New connection from 127.0.0.1:39986
2017-10-01 12:37:38,752 - INFO [kytos.core.tcp_server] (Thread-16) Connection lost with Client127.0.0.1:39980. Reason: Request closed by client.
2017-10-01 12:37:38,752 - INFO [kytos.core.tcp_server] (Thread-18) Connection lost with Client127.0.0.1:39984. Reason: Request closed by client.
2017-10-01 12:37:38,753 - INFO [kytos.core.tcp_server] (Thread-17) Connection lost with Client127.0.0.1:39982. Reason: Request closed by client.
2017-10-01 12:37:38,753 - INFO [kytos.core.tcp_server] (Thread-19) Connection lost with Client127.0.0.1:39986. Reason: Request closed by client.
2017-10-01 12:37:39,753 - INFO [kytos.core.tcp_server] (Thread-20) New connection from 127.0.0.1:39988
2017-10-01 12:37:39,754 - INFO [kytos.core.tcp_server] (Thread-21) New connection from 127.0.0.1:39990
2017-10-01 12:37:39,759 - INFO [kytos.core.tcp_server] (Thread-22) New connection from 127.0.0.1:39992
2017-10-01 12:37:39,760 - INFO [kytos.core.tcp_server] (Thread-23) New connection from 127.0.0.1:39994
```

- listening on port `8000`:

```
OF Address....: tcp://0.0.0.0:8000
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-01 13:32:41,751 - INFO [kytos.core.tcp_server] (Thread-8) Connection lost with Client 127.0.0.1:45574. Reason: Request closed by client.
2017-10-01 13:32:41,752 - INFO [kytos.core.tcp_server] (Thread-9) Connection lost with Client 127.0.0.1:45576. Reason: Request closed by client.
2017-10-01 13:32:42,752 - INFO [kytos.core.tcp_server] (Thread-14) New connection from 127.0.0.1:45578
2017-10-01 13:32:42,754 - INFO [kytos.core.tcp_server] (Thread-15) New connection from 127.0.0.1:45580
2017-10-01 13:32:43,751 - INFO [kytos.core.tcp_server] (Thread-14) Connection lost with Client127.0.0.1:45578. Reason: Request closed by client.
2017-10-01 13:32:43,752 - INFO [kytos.core.tcp_server] (Thread-15) Connection lost with Client127.0.0.1:45580. Reason: Request closed by client.
2017-10-01 13:32:44,752 - INFO [kytos.core.tcp_server] (Thread-16) New connection from 127.0.0.1:45582
2017-10-01 13:32:44,754 - INFO [kytos.core.tcp_server] (Thread-17) New connection from 127.0.0.1:45584
2017-10-01 13:32:45,752 - INFO [kytos.core.tcp_server] (Thread-16) Connection lost with Client127.0.0.1:45582. Reason: Request closed by client.
2017-10-01 13:32:45,752 - INFO [kytos.core.tcp_server] (Thread-17) Connection lost with Client127.0.0.1:45584. Reason: Request closed by client.
2017-10-01 13:32:46,752 - INFO [kytos.core.tcp_server] (Thread-18) New connection from 127.0.0.1:45586
2017-10-01 13:32:46,754 - INFO [kytos.core.tcp_server] (Thread-19) New connection from 127.0.0.1:45588

```

**With this patch**:

- listening on port `8000`:

```
OF Address....: tcp://0.0.0.0:8000
WEB UI........: http://0.0.0.0:8181/
2017-10-01 13:34:50,172 - INFO [root] (of_lldp) Running NApp: <Main(of_lldp, started 139771431077632)>
kytos $> 2017-10-01 13:34:50,753 - INFO [kytos.core.tcp_server] (Thread-12) New connection from 127.0.0.1:46084
2017-10-01 13:34:50,760 - INFO [kytos.core.tcp_server] (Thread-13) New connection from 127.0.0.1:46086
2017-10-01 13:34:50,776 - INFO [kytos/of_core] (Thread-18) Connection ('127.0.0.1', 46086), Switch 00:00:00:00:00:00:00:02: OPENFLOW HANDSHAKE COMPLETE
2017-10-01 13:34:50,776 - INFO [kytos/of_core] (Thread-18) Connection ('127.0.0.1', 46086), Switch 00:00:00:00:00:00:00:02: OPENFLOW HANDSHAKE COMPLETE
2017-10-01 13:34:50,786 - INFO [kytos/of_core] (Thread-23) Connection ('127.0.0.1', 46084), Switch 00:00:00:00:00:00:00:01: OPENFLOW HANDSHAKE COMPLETE
2017-10-01 13:34:50,786 - INFO [kytos/of_core] (Thread-23) Connection ('127.0.0.1', 46084), Switch 00:00:00:00:00:00:00:01: OPENFLOW HANDSHAKE COMPLETE
kytos $>
```

### DIFF
--- a/etc/kytos/kytos.conf.template
+++ b/etc/kytos/kytos.conf.template
@@ -32,6 +32,9 @@ listen = 0.0.0.0
 # openflow packets using TCP protocol. Default is 6633.
 port = 6633
 
+# Southbound protocol name of the TCP server. Don't use quotes in the string.
+protocol_name =
+
 # The api_port parameter tells kytos controller to expose a port to accept
 # incoming requests and to send a response from kytos API REST.
 # Default is 8181.
--- a/kytos/core/config.py
+++ b/kytos/core/config.py
@@ -67,6 +67,10 @@ class KytosConfig():
                             action='store',
                             help="Specify the working directory")
 
+        parser.add_argument('-s', '--protocol_name',
+                            action='store',
+                            help="Specify the southbound protocol")
+
         self.conf_parser, self.parser = conf_parser, parser
         self.parse_args()
 
@@ -85,6 +89,7 @@ class KytosConfig():
                         'listen': '0.0.0.0',
                         'port': 6633,
                         'foreground': False,
+                        'protocol_name': '',
                         'debug': False}
 
         """
@@ -101,6 +106,7 @@ class KytosConfig():
                     'listen': '0.0.0.0',
                     'port': 6633,
                     'foreground': False,
+                    'protocol_name': '',
                     'debug': False}
 
         options, argv = self.conf_parser.parse_known_args()
@@ -137,5 +143,6 @@ class KytosConfig():
         options.daemon = True if options.daemon in ['True', True] else False
         options.port = int(options.port)
         options.api_port = int(options.api_port)
+        options.protocol_name = str(options.protocol_name)
 
         return options
--- a/kytos/core/tcp_server.py
+++ b/kytos/core/tcp_server.py
@@ -24,7 +24,8 @@ class KytosServer(ThreadingMixIn, TCPServer):
     allow_reuse_address = True
     main_threads = {}
 
-    def __init__(self, server_address, RequestHandlerClass, controller):
+    def __init__(self, server_address, RequestHandlerClass, controller,
+                 protocol_name):
         """Create the object without starting the server.
 
         Args:
@@ -34,10 +35,12 @@ class KytosServer(ThreadingMixIn, TCPServer):
                 Class that will be instantiated to handle each request.
             controller (:class:`~kytos.core.controller.Controller`):
                 An instance of Kytos Controller class.
+            protocol_name (str): Southbound protocol name that will be used
         """
         super().__init__(server_address, RequestHandlerClass,
                          bind_and_activate=False)
         self.controller = controller
+        self.protocol_name = protocol_name
 
     def serve_forever(self, poll_interval=0.5):
         """Handle requests until an explicit shutdown() is called."""
@@ -65,7 +68,8 @@ class KytosRequestHandler(BaseRequestHandler):
     """
 
     known_ports = {
-        6633: 'openflow'
+        6633: 'openflow',
+        6653: 'openflow'
     }
 
     def __init__(self, request, client_address, server):
@@ -95,6 +99,9 @@ class KytosRequestHandler(BaseRequestHandler):
 
         self.connection = Connection(self.addr, self.port, self.request)
         server_port = self.server.server_address[1]
+        if self.server.protocol_name:
+            self.known_ports[server_port] = self.server.protocol_name
+
         if server_port in self.known_ports:
             protocol_name = self.known_ports[server_port]
         else:
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -186,7 +186,7 @@ class Controller(object):
         self.server = KytosServer((self.options.listen,
                                    int(self.options.port)),
                                   KytosRequestHandler,
-                                  self)
+                                  self, self.options.protocol_name)
 
         raw_event_handler = self.raw_event_handler
         msg_in_event_handler = self.msg_in_event_handler


cemsbr:
@viniciusarcanjo, thank you. I'll review this PR more deeply soon. As to the "connection lost" messages, I believe they were solved by kytos/of_lldp#5.

diraol (commented):
Thanks for your contribution @viniciusarcanjo !

Indeed we coded the TCP Server to allow other SB protocols, you got the idea right. =)

On the future we intend to remove it from kytos core (it may be a responsability of other NApps, which would allow multiple "TCPServers" running at the same time).

But, for now, your addition of 6653 port is welcome.

About the `protocol_name = 'openflow'` change, I think we can have another approach more well structured and less hardcoded by using the kytos config file.

At this moment we already support listening to other ports by changin the `port` setting on `/etc/kytos/kytos.conf` (I think this is how you are starting the tcp server on other ports).

Why not adding a new item over there called `tcp_server_protocol` and then pass it as an argument to `KytosServer` (https://github.com/kytos/kytos/blob/master/kytos/core/controller.py#L186) ?

What do you think about it?

viniarck:
@cemsbr I'm running all napps with the latest commits following master, including of_lldp (latest commit 4c384b1). I've cloned all official repos and symlinked them into kytos napps dir.

@diraol Exactly. that's how I'm doing. However, with the original else clause of this protocol_name variable, note that the OpenFlow handshake will never actually happen because the `"raw.in"` openflow event will never exist in the first place. Instead, `kytos/core."{server_port}".raw.in` will be. According to my tests, without this patch, all non-standard ports, including 6653 wouldn't work from the handshake perspective, even if you were to change the conf file or pass the argument via kytos cli `-P`.

I've rebased on top of master and incremented the code as you guys suggested. Thanks a lot! By the way, I came across this issue when I was running multiple controllers on containers at AmLight production environment. Basically, I needed Kytos to listen on other ports.

etc/kytos/kytos.conf.template
```diff
@@ -32,6 +32,9 @@ listen = 0.0.0.0
 # openflow packets using TCP protocol. Default is 6633.
 port = 6633
 
+# The southbound protocol that will be used by the TCP server
+protocol_name = openflow
```
diraol: @viniciusarcanjo we are trying to remove openflow 'dependency' from our core (we know that there is still some things to work on that sense). Because of that, I would rather prefer to leave the `'protocol_name'` default as an empty string. If someone is using the default openflow ports (6633 and 6653) this won't make any difference (considering this PR that improves our `known_ports` dict).

kytos/core/tcp_server.py
```diff
@@ -98,8 +102,10 @@ def setup(self):
         if server_port in self.known_ports:
             protocol_name = self.known_ports[server_port]
         else:
-            protocol_name = f'{server_port:04d}'
```
diraol: @viniciusarcanjo I would leave this as it is over here to allow the TCPServer  to work on any port even with an unkown southbound protocol.
I think it would be better to 'update' the 'known_ports' dictionary by doing, on the `KytosRequestHandler` `setup()` the following:

```python
def setup(self):
    self.addr = self.client_address[0]
    self.port = self.client_address[1]
    if self.server.protocol_name:
        self.known_ports.update((self.port, self.server.protocol_name))
```
Then the rest of the code can remain the same with the right fallback to using the port number on the event name if we do not know the protocol that is being used.

kytos/core/tcp_server.py
```diff
@@ -98,8 +102,10 @@ def setup(self):
...
-        self.connection.protocol.name = protocol_name
+            protocol_name = self.server.protocol_name
+        # strip() to make sure quotes won't mess up KytosEvent name
+        self.connection.protocol.name = str(protocol_name)\
```
diraol: Did you faced any problems with quotes?! This wouldn't be a problem, I think.

diraol (changes requested):

etc/kytos/kytos.conf.template
viniarck: Got it. I'll do.

kytos/core/tcp_server.py
viniarck: Good suggestion! I'll do this, except that this self.port is the port of the client (that came from this request), I believe you meant the port of the server, i.e., `server_port = self.server.server_address[1]`

kytos/core/tcp_server.py
viniarck: Kind of. This problem only shows up if the user sets the protocol_name in the cfg file with quotes or double quotes. However, if the user doesn't specify a string this isn't a problem. Since there are other strings that don't have quotes in the cfg file I believe the user should follow the same pattern and don't specify quotes.  For example @diraol, let's say:

If I were to set (in the kytos conf file):

`#` The southbound protocol that will be used by the TCP server (note the single quote bellow)
protocol_name = 'openflow'

and if I were to debug the protocol_name variable in the tcp_server, you'd see this, and it would be a problem:
```
2017-10-04 18:47:58,356 - INFO [kytos.core.tcp_server] (Thread-11) protocol_name 'openflow'
```

viniarck:
@diraol, I tried to get rid of the strip, however, it is really weird the way this configparser treats the empty string. Even if I were to convert the `options.protocol_name` to str, the empty string is not considered empty (for example `if options.protocol_name:` is considered `True` - mindblown), I debugged for hours but I didn't find a better alternative than using strip. Plus, this strip has the advantage of allowing the user to specify either single or double quotes in the protocol name, for instance, if the user sets the protocol as either "openflow", 'openflow' or openflow it will work. Nevertheless, if you guys don't like it, let me know. Thanks!

**Tests**

> Corsa port 8000, protocol_name = openflow

OF Address....: tcp://0.0.0.0:8000
WEB UI........: http://0.0.0.0:8181/
kytos $>

kytos $> 2017-10-05 00:39:40,709 - INFO [root] (kyslacker) Running NApp: <Main(kyslacker, started 139659292165888)>
2017-10-05 00:39:40,711 - INFO [root] (of_lldp) Running NApp: <Main(of_lldp, started 139659283773184)>
2017-10-05 00:39:40,716 - INFO [root] (kytos/of_core) Running NApp: <Main(kytos/of_core, started 139659275380480)>
2017-10-05 00:39:45,669 - INFO [kytos.core.tcp_server] (Thread-11) New connection from 67.17.206.198:39734
2017-10-05 00:39:46,190 - INFO [kytos/of_core] (Thread-15) Connection ('67.17.206.198', 39734), Switch 00:00:33:33:33:33:33:33: OPENFLOW HANDSHAKE COMPLETE

> Corsa + Dell port 6653, protocol_name = ''

OF Address....: tcp://0.0.0.0:6653
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-05 00:39:00,503 - INFO [root] (kyslacker) Running NApp: <Main(kyslacker, started 140188485129984)>
2017-10-05 00:39:00,519 - INFO [root] (of_lldp) Running NApp: <Main(of_lldp, started 140188000974592)>
2017-10-05 00:39:00,524 - INFO [root] (kytos/of_core) Running NApp: <Main(kytos/of_core, started 140187990013696)>
2017-10-05 00:39:00,612 - INFO [kytos.core.tcp_server] (Thread-10) New connection from 67.17.206.198:34610
2017-10-05 00:39:00,787 - INFO [kytos.core.tcp_server] (Thread-13) New connection from 67.17.206.198:34606
2017-10-05 00:39:00,834 - INFO [kytos.core.tcp_server] (Thread-16) New connection from 67.17.206.203:57229
2017-10-05 00:39:01,034 - INFO [kytos/of_core] (Thread-21) Connection ('67.17.206.198', 34610), Switch 00:00:33:33:33:33:33:32: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 00:39:01,343 - INFO [kytos/of_core] (Thread-23) Connection ('67.17.206.198', 34606), Switch 00:00:33:33:33:33:33:31: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 00:39:01,542 - INFO [kytos/of_core] (Thread-27) Connection ('67.17.206.203', 57229), Switch 00:02:34:17:eb:14:5c:00: OPENFLOW HANDSHAKE COMPLETE
kytos $>

kytos/core/tcp_server.py
diraol: Yes, you are right!! =)

diraol (changes requested):
@viniciusarcanjo indeed you are right about the "problem" with quotes on the configparser.

As we can see at [1], the default behavior of 'ini' files read by configparser is to consider all values as strings (`ini` is a `key: value` file). We had to do some 'casting' on the number items of the configs. So, if one set `"openflow"` as value, the quote will be considered as part of the string, not as a string delimiter.

It may be a little bit strange for python users, but I think that, as you mentioned, considering the other examples (paths and so on), we may expect users to understand that they may not use quotes.

To avoid any kind of trouble, I think that worth adding a comment just above the `protocol_name` line stating that the name should not be between quotes!

Also, the default value must be 'empty', not `''` as I suggested, since `''` will become the not empty string `"''"`.

All in all, this is a great contribution to us, thanks a lot!
As soon as you change the `protocol_name = ''` on the template I'll approve this PR. 

[1] https://docs.python.org/3/library/configparser.html#supported-ini-file-structure

viniarck:
@diraol First, thanks for this detailed explanation. I naively assumed this standard library would work in a "Pythonic" way :). I've just pushed again the changes according to your recommendations. Let me know if there's anything else pending. 

I ran again the same tests:

> Corsa and Dell port 6653, protocol_name = empty

```
OF Address....: tcp://0.0.0.0:6653
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-05 12:18:07,714 - INFO [kytos.core.tcp_server] (Thread-11) New connection from 67.17.206.203:52508
2017-10-05 12:18:08,243 - INFO [kytos/of_core] (Thread-15) Connection ('67.17.206.203', 52508), Switch 00:02:34:17:eb:14:5c:00: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 12:18:08,258 - INFO [kytos.core.tcp_server] (Thread-16) New connection from 67.17.206.198:37314
2017-10-05 12:18:08,260 - INFO [kytos.core.tcp_server] (Thread-17) New connection from 67.17.206.198:37318
2017-10-05 12:18:08,585 - INFO [kytos/of_core] (Thread-27) Connection ('67.17.206.198', 37314), Switch 00:00:33:33:33:33:33:31: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 12:18:08,586 - INFO [kytos/of_core] (Thread-28) Connection ('67.17.206.198', 37318), Switch 00:00:33:33:33:33:33:32: OPENFLOW HANDSHAKE COMPLETE
```

> Corsa port 8000, protcol_name = openflow

```
kytos $> 2017-10-05 12:18:40,016 - INFO [root] (kyslacker) Running NApp: <Main(kyslacker, started 140618822833920)>
2017-10-05 12:18:40,018 - INFO [root] (of_lldp) Running NApp: <Main(of_lldp, started 140618814179072)>
2017-10-05 12:18:40,023 - INFO [root] (kytos/of_core) Running NApp: <Main(kytos/of_core, started 140618797491968)>
2017-10-05 12:18:41,330 - INFO [kytos.core.tcp_server] (Thread-11) New connection from 67.17.206.198:42308
2017-10-05 12:18:41,700 - INFO [kytos/of_core] (Thread-15) Connection ('67.17.206.198', 42308), Switch 00:00:33:33:33:33:33:33: OPENFLOW HANDSHAKE COMPLETE
kytos $>
```

diraol:
@viniciusarcanjo I didn't know this behavior of `configparser` either. I've just learned it looking for after you described the problem here. Learning together! 

Just one last thing. I think I wasn't clear enough and I led you to a misunderstanding.

When I said to let the value as 'empty', I meant really empty, not the `empty` string:
`protocol_name = `
Sorry about it, after I read my own text I noticed that misleading information.

INI File Strucuture supports attributes without any value, and that was what I meant!

viniarck:
Got it. I just changed again. Now it looks right. Thanks :)

Tests are still passing:

```
OF Address....: tcp://0.0.0.0:6653
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-05 15:21:38,405 - INFO [kytos.core.tcp_server] (Thread-11) New connection from 67.17.206.198:51042
2017-10-05 15:21:38,412 - INFO [kytos.core.tcp_server] (Thread-12) New connection from 67.17.206.198:51046
2017-10-05 15:21:38,731 - INFO [kytos/of_core] (Thread-18) Connection ('67.17.206.198', 51046), Switch 00:00:33:33:33:33:33:32: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 15:21:38,740 - INFO [kytos/of_core] (Thread-20) Connection ('67.17.206.198', 51042), Switch 00:00:33:33:33:33:33:31: OPENFLOW HANDSHAKE COMPLETE
2017-10-05 15:21:47,541 - INFO [kytos.core.tcp_server] (Thread-37) New connection from 67.17.206.203:63877
2017-10-05 15:21:48,060 - INFO [kytos/of_core] (Thread-41) Connection ('67.17.206.203', 63877), Switch 00:02:34:17:eb:14:5c:00: OPENFLOW HANDSHAKE COMPLETE

```

```
OF Address....: tcp://0.0.0.0:8000
WEB UI........: http://0.0.0.0:8181/
kytos $> 2017-10-05 15:22:39,766 - INFO [kytos/of_core] (Thread-14) Connection ('67.17.206.198', 56352), Switch 00:00:33:33:33:33:33:33: OPENFLOW HANDSHAKE COMPLETE
```

diraol (approved):

kytos/core/tcp_server.py
```diff
@@ -95,6 +99,9 @@ def setup(self):
 
         self.connection = Connection(self.addr, self.port, self.request)
         server_port = self.server.server_address[1]
+        if self.server.protocol_name:
+            self.known_ports.update({server_port: self.server.protocol_name})
```
cemsbr: Maybe I'm being too pedantic here but, in this case, the line below is more common and readable while having the exactly same behavior as the line above.
```python
self.known_ports[server_port] = self.server.protocol_name
```

kytos/core/config.py
```diff
@@ -137,5 +143,8 @@ def _parse_options(self, argv):
         options.daemon = True if options.daemon in ['True', True] else False
         options.port = int(options.port)
         options.api_port = int(options.api_port)
+        options.protocol_name = str(options.protocol_name)
+        # options.protocol_name = '' if options.protocol_name in ['empty']\
+        #     else options.protocol_name
```
cemsbr: Could you please remove the comments?

cemsbr (changes requested):

cemsbr (commented):
Just a couple of changes.

kytos/core/tcp_server.py
viniarck: Definitely!

kytos/core/config.py
viniarck: My bad! Done!

cemsbr (approved):

PR #600, opened 2017-10-02
Allow any order for rest, class, static decorators

cemsbr:


renanrodrigo (approved):

diraol (approved):
@cemsbr any special reason for adding tests with `@rest` before `@staticmethod` and `@classmethod` but not after them?

PR #604, opened 2017-10-04
Change kytos daemon

macartur:


kytos/core/daemon.py
```diff
@@ -1,104 +1,32 @@
...
+
+    def run(self):
+        """Run the controller and keep alive."""
+        try:
+            self.controller.start()
+        except SystemExit as exception:
+            self.controller.log.error(exception)
+            self.controller.log.info("Kytos daemon start aborted.")
+        while True:
+            pass
```
diraol: Is this `while True` right? When is it going to be 'stopped'?
@cemsbr it looks strange to me, do you agree?

kytos/core/controller.py
```diff
@@ -123,7 +123,7 @@ def enable_logs(self):
     def start(self, restart=False):
         """Create pidfile and call start_controller method."""
         self.enable_logs()
-        if not restart:
+        if not restart and self.options.daemon is True:
```
diraol: `self.options.daemon is True` or `self.options.daemon is False`?

Isn't the `pidfile` automatically created by `KytosDaemon` and we need to manually create it when we are not on the `daemon` mode?

diraol (commented):
@macartur is there any relevant information that worth to be added to the CHANGELOG?

kytos/core/daemon.py
macartur: This is only to keep the process Alive. When the signal SIGTERM or SIGINT is received the process will call the controller.stop to stop the controller instance and after that, the process will be killed.

diraol: Ok! Thanks for the explanation!

diraol (dismissed):

kytos/core/daemon.py
cemsbr: @macartur, this busy waiting is using 100% of CPU.

cemsbr (changes requested):

beraldoleal:
Can you please, create a better description of this PR ? Or link to any issue ?

kytos/core/daemon.py
```diff
@@ -1,104 +1,35 @@
...
+        while True:
+            sleep(0.02)
+
+    def shutdown(self, signum):
+        """Stop the controller and shutdown the process."""
+        self.controller.stop()
+        while True:
+            if self.controller.status() == 'Stopped':
+                super().shutdown(signum)
+            sleep(0.02)
```
cemsbr: @macartur, if sleep is actually required, 0.5 is enough for the user to wait.

cemsbr (changes requested):

cemsbr (approved):

hdiogenes:
Needs a description and/or linked issue. Change what and why?

Looks like it conflicts with some changes in [PR 605](https://github.com/kytos/kytos/pull/605).

beraldoleal:
@macartur please provide a better description.

macartur:
The [PR 605](https://github.com/kytos/kytos/pull/605) has the same commits of this, so I will close this PR.

Issue #666, opened 2017-12-08
Create method to update web-ui


beraldoleal:
This method will download the new web-ui from a specific repository and update the `kytos/web-ui/` local files.

This should be able to be called from kytos console.

PR #668, opened 2017-12-12
Create method to update the web-ui

macartur:
This method can be called at the console or using utils via REST.
Fix kytos/kytos#666

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -3,8 +3,10 @@ import logging
 import os
 import sys
 import warnings
-from urllib.error import URLError
-from urllib.request import urlopen
+import zipfile
+from datetime import datetime
+from urllib.error import HTTPError, URLError
+from urllib.request import urlopen, urlretrieve
 
 from flask import Flask, request, send_from_directory
 from flask_cors import CORS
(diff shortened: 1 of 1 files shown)


renanrodrigo (approved):

kytos/core/api_server.py
```diff
@@ -118,6 +124,29 @@ def web_ui(self):
...
+        if not os.path.exists(self.flask_dir) or force:
+            # download from github
+            package = urlretrieve(uri)[0]
+            # remove files from web-ui/*
+            shutil.rmtree(self.flask_dir, ignore_errors=True)
+            os.mkdir(self.flask_dir)
+            # unzip and extract files to web-ui/*
+            zip_ref = zipfile.ZipFile(package, 'r')
+            zip_ref.extractall(self.flask_dir)
+            zip_ref.close()
```
beraldoleal: Here if the file is corrupted or something goes wrong, we lost the directory.

We don't need to use github api for now, but at least:

1. download zip file.
2. try unzip in a temp folder, for instance: `kytos-ui.latest`
3. move old folder to `kytos-ui.20171231`
4. move temp folder to kytos-ui

beraldoleal (changes requested):

kytos/core/api_server.py
```diff
@@ -72,6 +77,7 @@ def start_api(self):
         """
         self.register_core_endpoint('shutdown/', self.shutdown_api)
         self.register_core_endpoint('status/', self.status_api)
+        self.register_core_endpoint('update_ui/', self.update_web_ui)
```
beraldoleal: I would prefer like that: web/update

beraldoleal (changes requested):

PR #706, opened 2018-04-17
Change generated ui component name

macartur:
- The component name genereted will following the structure
  {username}-{nappname}-{component-section}-{filename}


kytos/core/api_server.py
```diff
@@ -147,12 +147,10 @@ def get_ui_components(self, filename):
         components = []
         for name in glob(path):
             group = name.split('/')
```
hdiogenes: I suggest making it clear in the code what is documented only on the commit message, as a comment or on the variables' names:
username, napp_name, component_section, filename = group.split('/')

beraldoleal: @macartur please do a double check if this suggestion is possible.

macartur: Fixed, I changed the variable name from `group` to `dirname`.

PR #709, opened 2018-04-18
Improve update_web_ui method

macartur:
- The method update_web_ui receive a version number.


kytos/core/api_server.py
```diff
@@ -161,15 +164,15 @@ def web_ui(self):
         """Serve the index.html page for the admin-ui."""
         return send_file(f"{self.flask_dir}/index.html")
 
-    def update_web_ui(self, force=True):
+    def update_web_ui(self, version='1.1.1', force=True):
```
beraldoleal: There is no github url for the latest zip ?

beraldoleal (changes requested):

PR #710, opened 2018-04-18
asyncio - new tcp server and controller

hdiogenes:
API Server, ipython, event handlers and event notifications are still
running on separate threads.

* New lib requirement: janus, for async+sync queues
* Now Kytos calls controller.stop() on SIGTERM
* Set daemon mode on threads created by @listen_to to avoiding locking
* Shorter names for modules on log messages

Known issues:
* No IPython history between sessions, to avoid SQLite error w/ threads

### DIFF
--- a/bin/kytosd
+++ b/bin/kytosd
@@ -1,5 +1,19 @@
 #!/usr/bin/env python3.6
+"""Start Kytos SDN Platform core."""
+
+import asyncio
+
 from kytos.core import kytosd
 
-if __name__ == "__main__":
-    kytosd.main()
+
+def main():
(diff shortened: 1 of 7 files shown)


kytos/web-ui/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
macartur: This file is not necessary, in the master we don't have this file anymore.

macartur (dismissed):

beraldoleal:
Nice work @hdiogenes. I have some minor comments, but we can discuss during the next days.

PR #722, opened 2018-05-08
Add available vlans

macartur:
Fix kytos/mef_eline#6

### DIFF
--- a/kytos/core/link.py
+++ b/kytos/core/link.py
@@ -8,6 +8,7 @@ import json
 from uuid import uuid4
 
 from kytos.core.common import GenericEntity
+from kytos.core.interface import TAGType
 
 
 class Link(GenericEntity):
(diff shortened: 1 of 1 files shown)


beraldoleal:
@macartur tests...

kytos/core/link.py
```diff
@@ -83,6 +84,19 @@ def make_tag_available(self, tag):
         else:
             return False
 
+    def available_vlans(self):
+        """Get all available vlans from each interface in the link."""
```
beraldoleal, line 88: Available_vlans on a link MUST be an intersection.

beraldoleal (changes requested):

kytos/core/link.py
macartur: Done.

PR #723, opened 2018-05-08
Add method to return the status of switch

macartur:


beraldoleal:
@macartur tests...

kytos/core/switch.py
```diff
@@ -107,6 +113,30 @@ def id(self):  # pylint: disable=invalid-name
...
+        elif self.is_administration_down():
+            return SwitchStatus.ADMINISTRATIVE_DOWN
+        else:
+            return SwitchStatus.OPERATION_DOWN
+
+    def is_administration_down(self):
+        """Verify if the switch is administration down.
+
+        Returns:
+            boolean: True if the switch is administration down, otherwise False
```
hdiogenes: On the top you use "administratIVE", here it is "administratION" three times. It should be "administrative" everywhere, right?

hdiogenes: On another note: is this method really necessary if all it does is return "not enabled"?

hdiogenes (changes requested):

kytos/core/switch.py
macartur: I guess this method is necessary to be used in another methods. I am only implementing the blueprint specification, this method is defined in: https://github.com/kytos/kytos/blob/master/docs/blueprints/EP013.rst

macartur: I am wrong with the "administratIVE" and "administratION", i will change the words to "administratIVE".

hdiogenes (approved):

macartur:
The issue kytos/mef_eline#5 that implements the [EP013](https://github.com/kytos/kytos/blob/master/docs/blueprints/EP013.rst) is already implemented in [kytos/core/common.py](https://github.com/kytos/kytos/blob/master/kytos/core/common.py#L9).I am closing the Pull request.

PR #727, opened 2018-05-10
Add methods as_dict as_json in UNI and TAG

macartur:


beraldoleal:
this is ok, but is out-of-date.

viniarck:
@beraldoleal why is this out-of-date? Is there a better way to serialize these classes via JSON? Thanks!

PR #732, opened 2018-05-13
Create the reload method to reload the napp code.

llgoncalves:


kytos/core/controller.py
```diff
@@ -703,3 +709,27 @@ def unload_napps(self):
...
+        napp_module = reload_module(napp_module)
+        self.napps[(username, napp_name)] = napp_module.Main(controller=self)
+
+    def rest_reload_napp(self, username, napp_name):
+        """Resquest reload a NApp."""
+        self.reload_napp(username, napp_name)
+        return 'reloaded', 200
+
+    def rest_reload_all_napps(self):
+        """Resquest reload all NApps."""
```
hdiogenes: typo: resquest (in two methods)

hdiogenes (changes requested):

kytos/core/controller.py
```diff
@@ -703,3 +709,27 @@ def unload_napps(self):
...
+        """Reload a NApp."""
+        try:
+            napp_module = self._import_napp(username, napp_name)
+        except ModuleNotFoundError as err:
+            self.log.error("Error loading NApp '%s/%s': %s",
+                           username, napp_name, err)
+            return
+
+        napp_module = reload_module(napp_module)
+        self.napps[(username, napp_name)] = napp_module.Main(controller=self)
```
hdiogenes: There's no indication that it worked successfully. Please add an INFO log line to Kytos console to tell that a napp was reloaded.

hdiogenes (changes requested):
Those `rest_*` methods are a good sign that we should refactor those methods in a different class, but this can be done in another PR.

hdiogenes:
As we discussed, there's a few more decisions to be made about this PR: when a NApp module is reloaded, should we run its init/setup methods again? 

I added the following code to kytos/of_core's `setup()` method. When I tried uninstalling/installing the napp again, I see the output on the console. When I just do a `kytos napps reload kytos/of_core`, I don't receive the log line.

```
diff --git a/main.py b/main.py
index 4b64552..6029c76 100644
--- a/main.py
+++ b/main.py
@@ -56,6 +56,9 @@ class Main(KytosNApp):
         self.of_core_version_utils = {0x01: of_core_v0x01_utils,
                                       0x04: of_core_v0x04_utils}
         self.execute_as_loop(settings.STATS_INTERVAL)
+        log.info('OpenFlow Core NApp.setup() called, obj: %s',
+                 id(self))
+

     def execute(self):
         """Method to be runned once on app 'start' or in a loop.
```

viniarck:
IMO, it's reasonable to assume that a NApp reload should run the setup method followed by the execute method.

PR #739, opened 2018-05-19
Added NApps System tests blueprint

viniarck:
Hi team @beraldoleal @macartur @hdiogenes @llgoncalves, 

I'd like to request your review on this blueprint, and figure out a milestone target. I feel like integration tests help us to speed up a lot, and even help with code review for PRs, if the PR is making integration tests to fail, then the developer just broke the NApp. It sucks to have to integration run tests manually (testing on your own with mininet, for example).

### DIFF
--- a/docs/blueprints/EP015.rst
+++ b/docs/blueprints/EP015.rst
@@ -1,0 +1,85 @@
+Summary
+=======
+
+Title
+-----
+System tests for NApps validation.
+
+Authors
+-------
+Vinicius Arcanjo (RNP/Kytos), Beraldo Leal (Kytos), Macatur Souza (Kytos), Humberto Diogenes (Kytos)
+
+Blueprint Status
+----------------
+work-in-progress
+
+Priority
+--------
+high
+
+Tags
+----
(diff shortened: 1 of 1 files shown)


viniarck:
Sometimes the boundaries of integration tests and system tests are quite blurry in the literature. But looks like what's proposed here is more close to what's considered system tests than integration tests, since the goal is to test as close to production as possible including actual networking devices. So, I've updated the blueprint. It's still WIP, eventually we'll bring this blueprint to a future release/dev sprint.

viniarck:
@beraldoleal @hdiogenes I'm removing the WIP from the tittle, since Jeronimo, @ajoaoff  and I have refined this blueprint. We've pushed a new prototype on pytest (https://github.com/amlight/flow_manager/tree/system_tests), I'll make sure scrutinizer can run it on docker containers too.  Soon enough, you guys can start reviewing the blueprint and the system tests results that we started with flow manager.

viniarck:
I just rebased, and set the discussion label.

hdiogenes (approved):

PR #745, opened 2018-06-06
Change generic entity to have active enabled methods

macartur:


kytos/core/common.py
```diff
@@ -22,9 +22,49 @@ def __init__(self):
         options = KytosConfig().options['daemon']
         self.metadata = {}
         # operational status with True or False
-        self.active = True
```
beraldoleal, line 25: Please, consider this approach:

```
GenericEntity:
    _active = False # Operational
    _enabled = False # Administrative
    
    def enable(self):
        self._enabled = True
    
    def disable(self):
        self._enabled = False
        
    def activate(self):
        self._active = True
    
    def deactivate(self):
        self._active = False
    
    def is_activate(self):
        return self._active
    
    def is_enabled(self):
        return self._enabled
        
Interface(GenericEntity):
    
Link(GenericEntity): 
   def is_enabled(self):
       return self.is_enabled() and self.endpoint_a.is_enabled() and self.endpoint_b.is_enabled()    

   def is_activate(self):
       return self.is_active() and self.endpoint_a.is_active() and self.endpoint_b.is_active()

Switch(GenericEntity):
    def enable(self):
        self._enabled = True

    def disable(self):
        for interface in self.interfaces:
            interface.disable()
        self._enabled = False 
```

beraldoleal (changes requested):

PR #762, opened 2018-06-28
Register user napps server

macartur:


docs/Makefile
```diff
@@ -77,7 +77,7 @@ clean:
 
 .PHONY: html
 html:
-	python home/make_home.py
+	python3.6 home/make_home.py
```
hdiogenes: Python 3.7 was released yesterday, so it would be nice to support it from now on. Replacing the hard-coded `python3.6` call with `$(PYTHON)` and a little boilerplate code at the top would be enough:

```
ifeq (${PYTHON}, )
    override PYTHON = python3.6
endif

all:
	echo $(PYTHON)
```

Then it would be just a matter of setting a python version when running make:

```
PYTHON=python3.7 make
```

hdiogenes (approved):

hdiogenes (approved):

PR #764, opened 2018-07-03
[Docs] Admin Guide / Installing

llgoncalves:
- Add the requirements.txt and  README.rst files to the Kytos documentation
- Write Admin Guide / Installing


docs/admin/installing.rst
```diff
@@ -2,7 +2,78 @@
...
+This will install Kytos with all the dependencies, like kytos-utils and
+python-openflow.
 
 
 Production Environment
 ======================
+
+To download now the latest release (it still a beta software) from our
+repository, first you need to clone *kytos* repository. So, install git
+before clone:
```
hdiogenes, line 52: 1. Instructions on how to install with Git should be moved to the dev guide, production environment should install with Pip.
2. Maybe we don't need to do this right now, but in the future we should have instructions on how to install kytos with a production web server.

hdiogenes (approved):

Issue #755, opened 2018-06-26
Exception when trying to read napps_pre_installed default value


hdiogenes:
It looks like the new version has a bug when there's no list of napps to pre-install. If I change the default value in the code to `'[]'` (str) instead of `[]` (list), it works.

```
git:(c215b56) kytosd -f
Exception in callback async_main() at /Users/humberto/src/kytos/kytos-master/kytos/core/kytosd.py:87
handle: <Handle async_main() at /Users/humberto/src/kytos/kytos-master/kytos/core/kytosd.py:87>
Traceback (most recent call last):
  File "/usr/local/python3.6-dbg/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/Users/humberto/src/kytos/kytos-master/kytos/core/kytosd.py", line 110, in async_main
    config = KytosConfig()
  File "/Users/humberto/src/kytos/kytos-master/kytos/core/config.py", line 79, in __init__
    self.parse_args()
  File "/Users/humberto/src/kytos/kytos-master/kytos/core/config.py", line 135, in parse_args
    self.options['daemon'] = self._parse_options(argv)
  File "/Users/humberto/src/kytos/kytos-master/kytos/core/config.py", line 157, in _parse_options
    options.napps_pre_installed = json.loads(options.napps_pre_installed)
  File "/usr/local/python3.6-dbg/lib/python3.6/json/__init__.py", line 348, in loads
    'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'list'
```

beraldoleal:
@macartur @hdiogenes lets try to fix this on rc2.

PR #766, opened 2018-07-04
napps_pre_installed bug with default value

macartur:
Fix kytos/kytos#755


hdiogenes (commented):
I'm a little worried that this approach may silently ignoring invalid values. Maybe we should just fix the default option value to the same format that is expected on the command-line? (str instead of list, or None?)

hdiogenes (approved):
Forget what I said, I got it now: if the value is already set it just won't mess with it.

PR #814, opened 2018-07-30
Add DELETE/PATCH EVCs

ajoaoff:


beraldoleal:
@ajoaoff thanks for this. Can you please rebase this pr? Is now out-of-date.

jab1982:
@ajoaoff , I think the operator should be able to change UNI_A and UNI_Z if needed. In fact, everything can be changed during the Patch process.

PR #819, opened 2018-08-21
Python 3.7 compatibility

hdiogenes:


hdiogenes:
It looks like some upgrade is breaking backwards-compatibility for Python 3.6 packaging, so we'll need to address that too.

PR #832, opened 2018-10-31
Support for meta-napps (fix #791)

### LINKED ISSUE #791: Update kytos napps loader in order to support meta NApp

hdiogenes:
* We assume that a napp without a `main` module is meta
* Enable all dependencies when enabling the main napp


beraldoleal:
Remember that we have to support enable and disable too.

Issue #877, opened 2019-02-13
Handle TypeError when the vlan_pool option is missing


gleybersonandrade:

beraldoleal:
@gleybersonandrade what option in which file? Please provide a better description.

gleybersonandrade:
Handle the TyperError Exception of the of_core's main file. This error ocurrs when the _vlan_pool_ value is missing in the Kytos configuration file (/etc/kytos/kytos.conf). In this exception, the JSON object must be str, bytes or bytearray, not dict.

``` Exception in thread Thread-2882:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/gleyberson/Developer/kytos-workspace/kytos/kytos/core/helpers.py", line 72, in threaded_handler
    handler(*args)
  File "/home/gleyberson/Developer/kytos-workspace/kytos-environment/var/lib/kytos/napps/kytos/of_core/main.py", line 112, in handle_features_reply
    switch = version_utils.handle_features_reply(self.controller, event)
  File "/home/gleyberson/Developer/kytos-workspace/kytos-environment/var/lib/kytos/napps/../napps/kytos/of_core/v0x04/utils.py", line 71, in handle_features_reply
    connection=connection)
  File "/home/gleyberson/Developer/kytos-workspace/kytos/kytos/core/controller.py", line 542, in get_switch_or_create
    self.set_switch_options(dpid=dpid)
  File "/home/gleyberson/Developer/kytos-workspace/kytos/kytos/core/controller.py", line 569, in set_switch_options
    vlan_pool = json.loads(self.options.vlan_pool)
  File "/usr/lib/python3.6/json/__init__.py", line 348, in loads
    'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'dict'

hdiogenes:
The fact that this is being triggered by `of_core` is a symptom of another problem: this option is being re-parsed every time a switch handshakes with Kytos. This doesn't necessarily mean that it re-reads `kytos.conf` and updates the in-memory value – it seems to be just re-parsing the value every time, which has no point. Maybe we should create another issue to address the refactoring of this code.

hdiogenes:
I've created issue #960 to address the refactoring suggested in my last comment.

PR #878, opened 2019-02-13
Handle the TypeError exception (#877)

gleybersonandrade:


hdiogenes (changes requested):
@gleybersonandrade, please update the commit message with a more meaningful one, preferably telling readers why you had to do this. Eg.: Avoid exception with missing vlan_pool (fix #877)

PR #881, opened 2019-02-19
Fix Linter errors

gleybersonandrade:
Fixed some linter errors in project

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -22,8 +22,8 @@ import re
 import sys
 import threading
 from concurrent.futures import ThreadPoolExecutor
-from importlib import import_module
 from importlib import reload as reload_module
+from importlib import import_module
 from importlib.util import module_from_spec, spec_from_file_location
 from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


hdiogenes (approved):

hdiogenes (approved):

kytos/core/controller.py
```diff
@@ -790,26 +789,33 @@ def unload_napps(self):
...
             self.unload_napp(username, napp_name)
 
-    def reload_napp(self, username, napp_name):
-        """Reload a NApp."""
-        self.unload_napp(username, napp_name)
-
-        mod_name = '.'.join(['napps', username, napp_name, 'main'])
+    def reload_napp_module(self, username, napp_name, napp_file):
+        """Reload a NApp Module."""
+        mod_name = '.'.join(['napps', username, napp_name, napp_file])
```
hdiogenes: This is a big change to be included together with linter fixes. Please separate this in another branch / PR.

hdiogenes (changes requested):
After a closer inspection, I noticed that the reload fix has some logic changes in the code that should be discussed separately, so I'm requesting that you move this commit to another PR.

hdiogenes (approved):

Issue #847, opened 2019-01-24
NApp reload is not getting new NApp settings


gleybersonandrade:
When I try to reload a NApp, the constants defined in the settings.py aren't update. I tried to install the loopnapp NApp, but when I changed the UPTIME_INTERVAL in settings.py and the Main Class setup method in main.py, only the second one was changed.

PR #882, opened 2019-02-28
Make NApp reload get new NApp settings (#847)

gleybersonandrade:
Fixed reload settings values in Napps

### DIFF
--- a/kytos/core/controller.py
+++ b/kytos/core/controller.py
@@ -796,24 +796,21 @@ class Controller:
             napp_module = import_module(mod_name)
         except ModuleNotFoundError as err:
             self.log.error("Module '%s' not found", mod_name)
-            return err
+            raise
         try:
             napp_module = reload_module(napp_module)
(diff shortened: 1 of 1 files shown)


kytos/core/controller.py
```diff
@@ -790,26 +789,33 @@ def unload_napps(self):
...
         try:
             napp_module = reload_module(napp_module)
-            self.log.info("NApp '%s/%s' successfully reloaded",
-                          username, napp_name)
         except ImportError as err:
             self.log.error("Error reloading NApp '%s/%s': %s",
                            username, napp_name, err)
-            return 400
+            return False
+        return True
```
hdiogenes: I think we shouldn't be using return values for error handling here. Maybe just re-raise the exception or create a new `NAppImportError` or something like that, and then handle that in the calling function.

hdiogenes (changes requested):

PR #891, opened 2019-03-20
Fix method get_next_available_tag (fix #883)

### LINKED ISSUE #883: Acquire same VLAN in concurrency test
I was trying to force an Error to acquire Link VLANs and I ended acquiring the same VLAN number for multiple requests.
First I used a [test_concurrently decorator](https://www.caktusgroup.com/blog/2009/05/26/testing-django-views-for-concurrency-issues/) to create a unit test.
I added a time.sleep(0.0001) between get_next_available_tag and use_tag to force a delay.
```
    def test_vlans(self):
        switch_a = Switch('00:00:00:00:00:01')
        switch_b = Switch('00:00:00:00:00:02')
        endpoint_a = Interface('eth0', 1, switch_a)
        endpoint_b = Interface('eth1', 2, switch_b)
        linkA = Link(endpoint_a, endpoint_b)

        @test_concurrently(10)
        def toggle_call_your_method_here():
            # perform the code you want to test here; it must be thread-safe
            vlan = linkA.get_next_available_tag()
            time.sleep(0.0001)
            linkA.use_tag(vlan)
            print('Link %s using %s vlan' % (linkA, vlan.value))

        toggle_call_your_method_here()
```

The result was:
```
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 1 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 2 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 3 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 4 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 4 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 5 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 5 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 6 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 5 vlan
Link <kytos.core.link.Link object at 0x7f9a3def3be0> using 6 vlan
```

rmotitsuki:
Fix the method get_next_available_tag to get and use the tag in single
 step. Deprecated the method use_tag, but maintained the behavior of the
 workflow using get tag / use tag.
Fix linter messages.

### DIFF
--- a/kytos/core/link.py
+++ b/kytos/core/link.py
@@ -92,7 +92,10 @@ class Link(GenericEntity):
                 self.endpoint_b.available_tags]
 
     def use_tag(self, tag):
-        """Remove a specific tag from available_tags if it is there."""
+        """Remove a specific tag from available_tags if it is there.
+
+        Deprecated: use only the get_next_available_tag method.
+        """
         if self.is_tag_available(tag):
             self.endpoint_a.use_tag(tag)
             self.endpoint_b.use_tag(tag)
@@ -106,9 +109,28 @@ class Link(GenericEntity):
 
     def get_next_available_tag(self):
         """Return the next available tag if exists."""
-        for tag in self.endpoint_a.available_tags:
-            if tag in self.endpoint_b.available_tags:
+        # Copy the available tags because in case of error
+        # we will remove and add elements to the available_tags
+        available_tags_a = self.endpoint_a.available_tags.copy()
+        available_tags_b = self.endpoint_b.available_tags.copy()
+
+        for tag in available_tags_a:
+            if tag in available_tags_b:
+                is_success = self.endpoint_a.use_tag(tag)
+                if not is_success:
+                    # Tag already in use. Try another tag.
+                    continue
+
+                is_success = self.endpoint_b.use_tag(tag)
+                if not is_success:
+                    # Tag already in use.
+                    # Return endpoint_a tag and try another tag.
+                    self.endpoint_a.make_tag_available(tag)
+                    continue
+
+                # Tag used successfully by both endpoints. Returning.
                 return tag
+
         return False
 
     def make_tag_available(self, tag):
--- a/kytos/core/link.py
+++ b/kytos/core/link.py
@@ -115,21 +115,21 @@ class Link(GenericEntity):
         available_tags_b = self.endpoint_b.available_tags.copy()
 
         for tag in available_tags_a:
-            if tag in available_tags_b:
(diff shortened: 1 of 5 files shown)


kytos/core/link.py
```diff
@@ -106,9 +109,28 @@ def is_tag_available(self, tag):
...
+        # we will remove and add elements to the available_tags
+        available_tags_a = self.endpoint_a.available_tags.copy()
+        available_tags_b = self.endpoint_b.available_tags.copy()
+
+        for tag in available_tags_a:
+            if tag in available_tags_b:
+                is_success = self.endpoint_a.use_tag(tag)
+                if not is_success:
+                    # Tag already in use. Try another tag.
+                    continue
```
hdiogenes: Suggestions:

1. Make the code more flat by removing it from the `if` clause. Just invert the condition: `if tag not in available_tags_b: continue`
1. I think the (repeated) temp variable could also be removed: `if not self.endpoint_a.use_tag(tag): continue`

kytos/core/link.py
```diff
@@ -106,9 +109,28 @@ def is_tag_available(self, tag):
...
+
+                is_success = self.endpoint_b.use_tag(tag)
+                if not is_success:
+                    # Tag already in use.
+                    # Return endpoint_a tag and try another tag.
+                    self.endpoint_a.make_tag_available(tag)
+                    continue
```
hdiogenes: Instead of using the word "Return" when no value is returned, I suggest changing the comment to "Mark the tag as available again", or something like that (it took me a few readings to understand why there was no `return` clause).

tests/helper.py
```diff
@@ -122,3 +123,50 @@ def new_handshaked_client(options=None):
...
+        def wrapper(*args, **kwargs):
+            """Thread wrapper."""
+            exceptions = []
+
+            def call_test_func():
+                """Call the test method."""
+                try:
+                    test_func(*args, **kwargs)
+                except Exception:
+                    raise
```
hdiogenes, line 160: Who is populating the `exceptions` list? Shouldn't that code be here instead of `raise`?

hdiogenes (changes requested):
I've added a few refactoring suggestions and a question about the new concurrent test helper.

hdiogenes (approved):

Issue #905, opened 2019-05-17
Change default OpenFlow port


gleybersonandrade:
The updated specs changed the port used by OpenFlow from 6633 to 6653. The first time this port was mentioned in an OpenFlow documentation was in version 1.3.4, but should be discontinued after version 1.4.0.

https://www.opennetworking.org/wp-content/uploads/2014/10/openflow-switch-v1.3.4.pdf
https://www.opennetworking.org/wp-content/uploads/2014/10/openflow-spec-v1.4.0.pdf

PR #906, opened 2019-05-17
Changed OpenFlow port to 6653 (#905)

gleybersonandrade:
Openflow version was changed from 6633 to 6653.


kytos/core/atcp_server.py
```diff
@@ -89,7 +89,6 @@ class KytosServerProtocol(asyncio.Protocol):
     """
 
     known_ports = {
-        6633: 'openflow',
```
beraldoleal: Here, I would suggest you to keep the old one too. There is no need to remove from here, now.

beraldoleal (changes requested):

PR #920, opened 2019-06-18
Send an event when NApp is done loading.

### LINKED ISSUE #919: NApps send an event after loading
After loading a NApp, an event can be launched to tell other interested NApps that this one has already loaded.

ajoaoff:


kytos/core/napps/base.py
```diff
@@ -241,6 +241,12 @@ def run(self):
 
         It should not be overriden.
         """
+        # Inform this NApp has been loaded
+        from kytos.core import KytosEvent
+        name = f'{self.username}/{self.name}.loaded'
+        event = KytosEvent(name=name, content={})
+        self.controller.buffers.app.put(event)
```
hdiogenes: 1. The import is here because it would case a circular import problem?
1. May these 4 lines could be refactored in a separate method.

ajoaoff: Yes, the import is there to avoid a circular import.. I'll put that code on a separate method.

hdiogenes:
@ajoaoff the linter is complaining about cyclic imports after updating from master.

PR #942, opened 2019-07-30
The user can now choose to enable or disable an interface to receive an LLDP packet.

cmagnobarbosa:
Now the user can choose to enable or disable a switch interface to receive LLDP packets.


jab1982:
This is good! Also, consider that it would be very good if the administrator could choose to enable or disable an interface to SEND an LLDP packet.

beraldoleal:
> This is good! Also, consider that it would be very good if the administrator could choose to enable or disable an interface to SEND an LLDP packet.

Yes, this will be made by of_lldp REST API interface.

PR #948, opened 2019-08-20
setup.py updated and run.txt added to standardize dep management

josemauro:
Now setup.py fetch dependencies from run.txt. The file "run.txt" it is genarated from "run.in" by pip-compile with all dependencies versions pinned.


kytos/web-ui/index.html
```diff
@@ -0,0 +1,18 @@
...
+  </head>
+  <body>
+    <div id="app"></div>
+    <script>
+      window.kytos_server = location.protocol + "//" + location.host + "/";
+      window.kytos_server_api = location.protocol + "//" + location.host + "/api/";
+    </script>
+    <script src="/dist/build.js"></script>
+  </body>
+</html>
```
hdiogenes: Is this file here by accident?

josemauro: yes, I'm removing it

Issue #939, opened 2019-07-24
Improve logging when error messages arrive from the network


beraldoleal:
Today, some OFP_ERROR's are being logged at the DEBUG log level. We need to migrate this to ERROR. Only code-related errors must be in DEBUG.

We also need to create the ability to change the verbosity level during runtime.

beraldoleal:
Related #933 .

PR #956, opened 2019-09-11
Modify loggers to begin the hierarchy name with "kytos"

rmotitsuki:


rmotitsuki:
Related to issue #939

PR #966, opened 2019-10-18
Create Authentication module

gleybersonandrade:


ajoaoff (changes requested):
The methods to manipulate the user database should themselves be used only by authenticated users. There should be a method to create an initial user on deploy, or through CLI on the server.

ajoaoff (changes requested):
I believe `_list_users` and `_list_user` should also require an authenticated user.
The method to create the default user should not be called by `__init__` , because the default user will be recreated every time Kytos is started.

hdiogenes (approved):

kytos/core/auth.py
```diff
@@ -0,0 +1,334 @@
...
+            if error or not box:
+                response = {
+                    "answer": "User data cannot be shown",
+                    "code": HTTPStatus.INTERNAL_SERVER_ERROR.value,
+                }
+            else:
+                response = {
+                    "answer": {"data": box.data},
+                    "code": HTTPStatus.OK.value,
+                }
```
ajoaoff: An empty `box` is not an error, it means there is no user with the given uid.

ajoaoff: An empty `box` is not an error, it means there is no user with the given uid.

kytos/core/auth.py
```diff
@@ -0,0 +1,334 @@
...
+            result = f"Incorrect username or password: {exc}"
+            return result, HTTPStatus.UNAUTHORIZED.value
+
+    def _find_user(self, uid):
+        """Find a specific user using Storehouse."""
+        response = {}
+
+        def _find_user_callback(_event, box, error):
+            nonlocal response
+            if error or not box:
```
ajoaoff:
```suggestion
            if error:
```

kytos/core/auth.py
```diff
@@ -0,0 +1,334 @@
...
+                response = {
+                    "answer": "User data cannot be shown",
+                    "code": HTTPStatus.INTERNAL_SERVER_ERROR.value,
+                }
+            else:
```
ajoaoff, line 174:
```suggestion
            elif not box:
                response = {
                    "answer": f'User with uid {uid} not found',
                    "code": HTTPStatus.NOT_FOUND.value
                }
            else:
```

kytos/core/auth.py
```diff
@@ -0,0 +1,334 @@
...
+            if error or not box:
+                response = {
+                    "answer": "User has not been deleted",
+                    "code": HTTPStatus.INTERNAL_SERVER_ERROR.value,
+                }
+            else:
+                response = {
+                    "answer": "User successfully deleted",
+                    "code": HTTPStatus.OK.value,
+                }
```
ajoaoff: Same as `_find_user`. Here maybe it would be better to check first if the user exists.

ajoaoff (changes requested):

ajoaoff (approved):

kytos/core/config.py
```diff
@@ -76,6 +76,10 @@ def __init__(self):
                             action='store_true',
                             help="Enable all new Entities by default.")
 
+        parser.add_argument('-C', '--create_superuser',
```
hdiogenes:
```suggestion
        parser.add_argument('-C', '--create-superuser',
```

PR #974, opened 2019-11-01
Fix config path for install kytos without virtual env

rmotitsuki:
Fix setup that was creating kytos configuration files with relative paths.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -156,7 +156,8 @@ class CommonInstall:
         """
         from jinja2 import Template
 
-        kwargs['prefix'] = Path(str(kwargs['prefix']).rstrip('/'))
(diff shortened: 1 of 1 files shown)


josemauro:
After reinstall kytos (without a virtual env) using "python setup.py develop"  and trying to run using "kytosd -f" I get the following message:
"Log config file "./etc/kytos/logging.ini" does not exist. Using default Python logging configuration.
"'. 
We can see that kytos is looking for "loggin.ini" from the current directory "./" and not finding it
Screenshot from 2019-11-04 17-39-46
.

josemauro (commented):
The kytos installation works fine with this modifications.

hdiogenes (approved):

PR #984, opened 2019-12-03
Teach users how to protect an endpoint: Add authentication…

cmagnobarbosa:
Now that kytos will support secure endpoints, we have to create a documentation explaining the new functionalities and a tutorial teaching users how to protect endpoints.

### DIFF
--- a/docs/developer/auth.rst
+++ b/docs/developer/auth.rst
@@ -1,0 +1,92 @@
+******************
+Auth Documentation
+******************
+
+The authentication module (kytos.core.authentication) is a resource under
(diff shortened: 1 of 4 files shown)


hdiogenes (changes requested):
1. Respect 80 char limit
1. s/$curl/$ curl/
1. standardize indentation on code-blocks
1. substitute all 'URL's with the complete URL example to make it easier for users to copy/paste from the documentation
1. add spaces after ":" on HTTP request headers, dictionaries, JSON, etc.

Issue #1007, opened 2019-12-16
Authentication - Create a configuration field to change token expiration time


rmotitsuki:
Token expiration time is hardcoded for 10 minutes.
https://github.com/kytos/kytos/blob/master/kytos/core/auth.py#L150
It should be inside kytos.conf

cmagnobarbosa:
Related #1009

gleybersonandrade:
Related #1128

PR #1010, opened 2019-12-17
Increase token expiration time

cmagnobarbosa:
Today the token expires  quickly(10 minutes). This commit increased token expiration time to 180 minutes.
In future this time can be configured by user. Related: #1007

### DIFF
--- a/kytos/core/auth.py
+++ b/kytos/core/auth.py
@@ -18,6 +18,7 @@ __all__ = ['authenticated']
 LOG = logging.getLogger(__name__)
 TOKEN_EXPIRATION_MINUTES = 180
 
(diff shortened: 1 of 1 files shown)


kytos/core/auth.py
```diff
@@ -147,7 +147,7 @@ def _authenticate_user(self):
             if user.get("password") != hashlib.sha512(password).hexdigest():
                 raise KeyError
             time_exp = datetime.datetime.utcnow() + datetime.timedelta(
-                minutes=10
+                minutes=180
```
hdiogenes: Declare global constant, avoid magic number

PR #1044, opened 2020-03-25
Move the creation of config files to run time 

### LINKED ISSUE #1042: Create configuration only in post-install (or kytosd bootstrap)
Today Kytos tries to create its configuration at install time, a behavior that causes three problems for us:

* `SandboxViolation` when installing Kytos as a dependency from PyPI - #494
* Distribution only in source format (no wheel file) - #922
* "There is no config file." when starting `kytosd` - #951

That means we have to change the creation of configuration files to `post-install` or to the `kytosd` initialization so that we can close those three issues.

Older issues also related: #471 #385.

josemauro:
Moving the creation of config files to run time instead installation time. 
- The 'config.py' create a folder 'etc/kytos' and 'kytos.conf' if the file is not found.
- The 'kytosd.py' create 'var/run' and  folders.


kytos/core/config.py
```diff
@@ -125,18 +130,55 @@ def parse_args(self):
...
+        kwargs['jwt_secret'] = uuid.uuid4().hex
+
+        # Create the paths used by Kytos.
+        directories = [os.path.join(BASE_ENV, 'etc/kytos')]
+        for directory in directories:
+            if not os.path.exists(directory):
+                os.makedirs(directory)
+
+        for path in templates:
+            with open(path, 'r', encoding='utf-8') as src_file:
```
cmagnobarbosa: Running this modification, the error FileNotFound is displayed.

The error message:

FileNotFoundError: [Errno 2] No such file or directory: 'etc/kytos/kytos.conf.template'

To reproduce this error:

- Create a new virtual environment.
- Clone this repository: git clone git@github.com:kytos/kytos.git
- Get this modification: git fetch origin refs/pull/1044/head
- make checkout for this new branch: git checkout -b josemauro/fix-issue-1042 FETCH_HEAD
- Install python-openflow, kytos-utils, kytos.
- Run kytos: kytosd -f

cmagnobarbosa (changes requested):

cmagnobarbosa (approved):

kytos/core/kytosd.py
```diff
@@ -23,6 +29,18 @@ def in_prompt_tokens(self):
...
+class CommonInstall:
+    """Class with common methods used by children classes."""
+
+    @staticmethod
+    def create_pid_folder():
+        """Create the folder in /var/run to hold the pidfile."""
+        pid_folder = os.path.join(BASE_ENV, 'var/run/kytos')
+        os.makedirs(pid_folder, exist_ok=True)
+        if BASE_ENV == '/':  # system install
+            os.chmod(pid_folder, 0o1777)  # permissions like /tmp
```
hdiogenes: No need for class + static method when all you're going to use is a function.

kytos/core/config.py
```diff
@@ -125,18 +131,60 @@ def parse_args(self):
...
+        config.read(config_file)
+
+        defaults.update(dict(config.items("daemon")))
 
         self.parser.set_defaults(**defaults)
 
         self.options['daemon'] = self._parse_options(argv)
 
+    @staticmethod
+    def _generate_file_from_template(templates,
```
hdiogenes: `_generate_file_from_template` can be moved to a function instead of an `staticmethod`

kytos/core/config.py
```diff
@@ -125,18 +131,60 @@ def parse_args(self):
...
+
+        If no destination is passed, the new conf file will be created on the
+        directory of the template file.
+
+        Args:
+            template (string):    Path of the template file
+            destination (string): Directory in which the config file will
+                                  be placed.
+        """
+        from jinja2 import Template
```
hdiogenes: Move import to top of file.

hdiogenes (changes requested):

kytos/core/config.py
```diff
@@ -147,44 +149,6 @@ def parse_args(self):
 
         self.options['daemon'] = self._parse_options(argv)
 
-    @staticmethod
-    def _generate_file_from_template(templates,
-                                     destination=Path(__file__).parent,
-                                     **kwargs):
```
hdiogenes: It looks like this function name should be in the plural form: `_generate_files_from_templates()`, right?
My suggestion: `_render_config_templates()`

josemauro: I liked the suggestion and already applied it.

kytos/core/kytosd.py
```diff
@@ -84,6 +98,12 @@ def start_shell(controller=None):
 
 def main():
     """Read config and start Kytos in foreground or daemon mode."""
+    # data_files is not enough when installing from PyPI
+    for file in ETC_FILES:
+        shutil.copy2(file, Path(BASE_ENV) / file)
```
hdiogenes: What's the purpose of this when ETC_FILES is just an empty list?

setup.py
```diff
@@ -138,60 +132,56 @@ def run(self):
...
+#         if BASE_ENV == '/':  # system install
+#             os.chmod(pid_folder, 0o1777)  # permissions like /tmp
+
+#     @staticmethod
+#     def create_paths():
+#         """Create the paths used by Kytos in develop mode."""
+#         directories = [os.path.join(BASE_ENV, 'etc/kytos')]
+#         for directory in directories:
+#             if not os.path.exists(directory):
+#                 os.makedirs(directory)
```
hdiogenes: The commented lines can be removed.

setup.py
```diff
@@ -205,61 +195,19 @@ def run(self):
...
+#    """Recommended setup for developers.
+#
+#    The following feature are temporarily remove from code:
+#    Instead of copying the files to the expected directories, a symlink is
+#    created on the system aiming the current source code.
+#    """
+#
+#    def run(self):
+#        """Install the package in a developer mode."""
+#        super().run()
```
hdiogenes, line 162: remove commented code

setup.py
josemauro: It's done.

kytos/core/kytosd.py
josemauro: I'm not sure, but It's something that already were in the project, I'll investigate it to understand. For now, I don't see any reason to keep that.

setup.py
josemauro: It's done.

kytos/core/kytosd.py
josemauro: It's possible to check that the ETC_FILES got an empty list in commit c4b1a1a9, and stayed empty since then. I just removed this code and tested the installation. Everything seems to work fine.

kytos/core/kytosd.py
```diff
@@ -23,6 +27,14 @@ def in_prompt_tokens(self):
         return [(Token.Prompt, 'kytos $> ')]
 
 
+def _create_pid_folder():
+    """Create the folder in /var/run to hold the pidfile."""
+    pid_folder = os.path.join(BASE_ENV, 'var/run/kytos')
+    os.makedirs(pid_folder, exist_ok=True)
+    if BASE_ENV == '/':  # system install
+        os.chmod(pid_folder, 0o1777)  # permissions like /tmp
```
hdiogenes: `s/folder/directory/` (or `dir`)

josemauro: I replaced it. Thanks.

Issue #1043, opened 2020-03-25
Create "from_dict" methods


ajoaoff:
Some classes, like `Interface`, do not have a `from_dict` method. Because of that, NApps like `kytos/mef_eline` has its own methods to create those classes from a dictionary. That should not be the case, as other NApps who need that would need to implement it too.

PR #1045, opened 2020-03-30
Create from_dict method for Interface, UNI, Switch and Link

cmagnobarbosa:
Some classes, like Interface, UNI, Switch and Link do not have a from_dict method. This commit solves this problem.

Related https://github.com/kytos/kytos/issues/1043

### DIFF
--- a/kytos/core/interface.py
+++ b/kytos/core/interface.py
@@ -418,6 +418,17 @@ class Interface(GenericEntity):  # pylint: disable=too-many-instance-attributes
             iface_dict['stats'] = self.stats.as_dict()
         return iface_dict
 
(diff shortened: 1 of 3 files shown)


gleybersonandrade:
Related: https://github.com/kytos/kytos/issues/1043

hdiogenes (changes requested):
As we've discussed in the last meeting, this approach will have the downside of creating many instances of the same `Interface` object (with different values for the `available_tags` attribute, for example).

On the other hand, this is the approach which is already being done in some NApps, so at least we can centralize the logic here, and when we fix this it'll be fixed for all NApps that depend upon it.

PR #1056, opened 2020-04-13
Fix new linter issues

### LINKED ISSUE #1055: Solve linter issues raised by the new version of Pylint
Now, kytos is using a new version of Pylint and some new linter issues were raised, so it has been disabled. We have to solve these linter issues.

gleybersonandrade:
There were some linter issues raised by new Pylint version. This commit fixes all of them.

### DIFF
--- a/kytos/core/napps/__init__.py
+++ b/kytos/core/napps/__init__.py
@@ -1,8 +1,8 @@
 """Module responsible for running and managing NApps."""
+from kytos.core.api_server import APIServer
+
 from .base import KytosNApp, NApp
-from .manager import NAppsManager
-# pylint does not recognize the import below as local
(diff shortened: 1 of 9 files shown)


kytos/core/napps/__init__.py
```diff
@@ -1,8 +1,8 @@
 """Module responsible for running and managing NApps."""
+from kytos.core.api_server import APIServer
+
 from .base import KytosNApp, NApp
-from .manager import NAppsManager
-# pylint does not recognize the import below as local
-from kytos.core.api_server import APIServer  # pylint: disable=C0411
+from .manager import NAppsManager  # pylint: disable=cyclic-import
```
hdiogenes, line 5: Is this warning a false positive?

Issue #1057, opened 2020-04-15
NApp does not reload all submodules in runtime


cmagnobarbosa:
Today, the `napp_reload` function doesn't load all submodules related to an NApp, only the main.py and the settings.py are reloaded.
This bug results in problems in the NApp to load changes and new dependencies at runtime.

reload_napp code:
``` python
def reload_napp(self, username, napp_name):
        """Reload a NApp."""
        self.unload_napp(username, napp_name)
        try:
            self.reload_napp_module(username, napp_name, 'settings')
            self.reload_napp_module(username, napp_name, 'main')
        except (ModuleNotFoundError, ImportError):
            return 400
        self.log.info("NApp '%s/%s' successfully reloaded",
                      username, napp_name)
        self.load_napp(username, napp_name)
        return 200
```

Related issues :
- https://github.com/kytos/kytos/issues/979

Mentioned in:

- https://github.com/kytos/kytos/issues/847
- https://github.com/kytos/kytos/issues/834

PR #1058, opened 2020-04-18
Now reload_napp reload all NApp submodules

cmagnobarbosa:
Now all submodules the NApps are reloaded.

Related https://github.com/kytos/kytos/issues/1057


hdiogenes:
We'll probably create a new command to `restart` NApps and leave the `reload` command to just reload settings, as it is expected from Linux daemons.

PR #1061, opened 2020-05-05
Create lib with helpers to create tests

gleybersonandrade:
Today, each NApp needs to implement its own helpers.py file, and this code is sometimes duplicated. This commit creates a lib on kytos with helpers.py to be used as a utility in all tests.

### DIFF
--- a/kytos/lib/helpers.py
+++ b/kytos/lib/helpers.py
@@ -1,0 +1,80 @@
+"""Module with utilities to create tests."""
+from unittest.mock import MagicMock, Mock, create_autospec
+
+from kytos.core import Controller
+from kytos.core.config import KytosConfig
+from kytos.core.connection import (Connection, ConnectionProtocol,
+                                   ConnectionState)
+from kytos.core.events import KytosEvent
+from kytos.core.interface import Interface
+from kytos.core.link import Link
+from kytos.core.switch import Switch
+
+
+def get_controller_mock():
+    """Return a controller mock."""
+    options = KytosConfig().options['daemon']
+    controller = Controller(options)
+    controller.log = Mock()
+    return controller
+
+
+def get_interface_mock(name, port_number, switch, address="00:00:00:00:00:00"):
+    """Return a interface mock."""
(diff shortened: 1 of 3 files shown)


kytos/lib/helpers.py
```diff
@@ -0,0 +1,113 @@
...
+def get_interface_mock(name, port_number, switch):
+    """Return a interface mock."""
+    interface = create_autospec(Interface)
+    interface.id = "{}:{}".format(switch.dpid, port_number)
+    interface.name = name
+    interface.port_number = port_number
+    interface.switch = switch
+    interface.address = "00:00:00:00:00:00"
+    interface.lldp = True
+    return interface
```
ajoaoff: MAC address could also be a parameter, with a default of `00:00:00:00:00:00`. I think the method is more useful when you can set that field.

kytos/lib/helpers.py
```diff
@@ -0,0 +1,113 @@
...
+    link_1 = get_link_mock(interface_a1, interface_b1)
+    link_2 = get_link_mock(interface_a2, interface_c1)
+    link_3 = get_link_mock(interface_b2, interface_c2)
+
+    topology = MagicMock()
+    topology.links = {"1": link_1, "2": link_2, "3": link_3}
+    topology.switches = {switch_a.dpid: switch_a,
+                         switch_b.dpid: switch_b,
+                         switch_c.dpid: switch_c}
+    return topology
```
ajoaoff: I'm not sure if putting this mock here is correct, since `Topology` is an entity unknown to Kytos Core.

ajoaoff (changes requested):

kytos/lib/helpers.py
gleybersonandrade: Thanks @ajoaoff, I'll create this parameter.

kytos/lib/helpers.py
gleybersonandrade: Sure @ajoaoff, maybe we can put this mock only in NApps helpers, I'll move it. Thanks!

PR #1080, opened 2020-06-18
Test if UNI has tag in as_dict.

### LINKED ISSUE #1079: UNI as_dict method assumes user_tag exists
The `user_tag` attribute on `UNI` is optional, but the method `as_dict` assumes it exists and, when calling `self.user_tag.as_dict()` an exception is raised if `user_tag` is `None`.

ajoaoff:
As tag is optional in UNI, the method as_dict has to consider the case
where it is None. Now, the key 'user_tag' in the dict is created only
if user_tag attribute exists. Fix #1079.

### DIFF
--- a/kytos/core/interface.py
+++ b/kytos/core/interface.py
@@ -450,10 +450,10 @@ class UNI:
 
     def as_dict(self):
         """Return a dict representating a UNI object."""
-        uni_dict = {'interface_id': self.interface.id}
-        if self.user_tag:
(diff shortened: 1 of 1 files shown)


gleybersonandrade (approved):

italovalcy (approved):
I've reviewed the PR, and it sounds very straight-forward and correct. It fixes the issue when the requestor does not provide and TAG value.

PR #1087, opened 2020-07-06
Method to handle HTTPExceptions

### LINKED ISSUE #1086: Use Flask's error handler with HTTP errors
When something happens that lead to a reply with an HTTP error (a bad request, for instance), the application itself has to format the error message and return it with the correct HTTP error code.
Using Flask's error handler, the application can only raise the appropriate exception and, if the error comes from an external call (like `request.get_json()`), the application may do nothing.

ajoaoff:
New method to handle HTTPException now return a JSON with
error code, name and description. Now NApps can simple raise a
HTTPException setting a description to return in case of error.
Fix #1086.

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -15,6 +15,7 @@ from urllib.request import urlopen, urlretrieve
 from flask import Flask, jsonify, request, send_file
 from flask_cors import CORS
 from flask_socketio import SocketIO, join_room, leave_room
+from werkzeug.exceptions import HTTPException
 
 
 class APIServer:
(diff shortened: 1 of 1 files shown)


kytos/core/api_server.py
```diff
@@ -61,6 +62,22 @@ def __init__(self, app_name, listen='0.0.0.0', port=8181,
         # Update web-ui if necessary
         self.update_web_ui(force=False)
 
+        @self.app.errorhandler(HTTPException)
+        def handle_exception(e):
```
cmagnobarbosa: You can maybe change the variable name from `e` to a more explicit name as `http_error` or `err`. This approach avoids the warning of the invalid-name.

ajoaoff: Variable name changed.

cmagnobarbosa (approved):

josemauro (approved):

Issue #960, opened 2019-10-02
Refactor `Controller.set_switch_options`: use config cache


hdiogenes:
In #877, we noticed that when `of_core` calls the Kytos controller to instantiate a new switch, the controller re-parses `kytos.conf` to get the `vlan_pool` value again. This is not necessary - the configuration should already be loaded in memory, not parsed again. 

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "~/kytos/kytos/core/helpers.py", line 72, in threaded_handler
    handler(*args)
  File "~/venv/var/lib/kytos/napps/kytos/of_core/main.py", line 112, in handle_features_reply
    switch = version_utils.handle_features_reply(self.controller, event)
  File "~/venv/var/lib/kytos/napps/../napps/kytos/of_core/v0x04/utils.py", line 71, in handle_features_reply
    connection=connection)
  File "~/kytos/kytos/core/controller.py", line 542, in get_switch_or_create
    self.set_switch_options(dpid=dpid)
  File "~/kytos/kytos/core/controller.py", line 569, in set_switch_options
    vlan_pool = json.loads(self.options.vlan_pool)
  File "/usr/lib/python3.6/json/__init__.py", line 348, in loads
    'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'dict' 
```

hdiogenes:
The relevant lines in the traceback above are these:

```
  File "~/kytos/kytos/core/controller.py", line 569, in set_switch_options
    vlan_pool = json.loads(self.options.vlan_pool)
```

In the same `controller` file, the `self.options` attribute definition has this comment:
```
        #: Object generated by ParseArgs on config.py file
        self.options = options
```

Which means that the `json.loads` operation should probably be moved to `config.KytosConfig.parse_args`.

PR #1101, opened 2020-07-23
Fix for issue #960

slinksoft:
The Capstone 1 team at FIU attempted to solve issue #960 . With Humberto's advice, we move the json.loads call from controller.py to config.py, so that it is only parsed once and not re-parsed in set_switch_options. In controller.py, it is now loaded from memory.

If there is a mistake in this fix, please let us know, and we will address it ASAP. Also, due to our lack of understanding of using git and Github at the beginning of the semester, a lot of commits are shown. Commits on date 7/22/2020 are relevant to this pull request.


.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,6 @@
+{
+  "ExpandedNodes": [
+    ""
+  ],
+  "PreviewInSolutionExplorer": false
+}
```
hdiogenes, line 6: Please remove all the `.vs` files from the PR. You can add `.vs` to your personal `.gitignore`, or make another PR adding it to kytos' project-level `.gitignore`.

kytos/core/controller.py
```diff
@@ -128,6 +129,27 @@ def __init__(self, options=None, loop=None):
...
+        traceback.print_exception(exctype, value, tb)
+        print(tb)
+        # Configure to log uncaught exceptions to errlog file
+        # pylint: disable=logging-format-interpolation
+        logging.basicConfig(filename='kytos/kytos/core/errlog.log',
+                            format='%(asctime)s:%(pathname)'
+                            's:%(levelname)s:%(message)s')
+        print('Uncaught Exception: {0}'.format(str(value)))
+        # pylint: disable=logging-format-interpolation
+        logging.exception('Uncaught Exception: {0}'.format(str(value)))
```
hdiogenes, line 152: I think this code is from another PR, right? Maybe you should create a separate branch without this code.

kytos/core/config.py
```diff
@@ -145,6 +145,10 @@ def parse_args(self):
 
         defaults.update(dict(config.items("daemon")))
 
+        # Gets vlan_pool after default is updated with values
+        # from config gile. then parses vlan_pool via json.loads
+        defaults.update(vlan_pool=json.loads(defaults.get('vlan_pool')))
```
hdiogenes, line 150: A final check that needs to be made is if the "vlan_pool" value is really being read when defined in `kytos.conf`.

hdiogenes (changes requested):

PR #1141, opened 2020-08-26
Add authentication to REST methods based on configuration option

### LINKED ISSUE #1136: Make authentication user configurable
Authentication of REST methods is configured by the NApp developer, who defines which methods are authenticated decorating those methods with `@authenticated`.
It is desirable that the user can do that configuration and define which methods should be authenticated or not.
Related to #1128.

ajoaoff:
Users can configure kytos to add configuration to REST methods. This PR does the following:

- remove unecessary `rest` import in `kytos.core.helpers`;
- add configuration option `authenticate_urls`;
- decorate defined methods with `authenticated` before registering in Flask.

The configuration is based on URLs, so if any URL a method is registered to matches `authenticate_urls`, the method will
enforce authentication.
An endpoint URL is considered to match `authenticate_urls` if any URL defined there is a substring of the endpoint URL.
For example, assume `authenticate_urls = ["kytos/mef_eline", "pathfinder"]`, then:
- `/api/kytos/mef_eline/1` matches;
- `/api/kytos/pathfinder` matches;
- `/api/amlight/pathfinder` matches.
- `/api/amlight/mef_eline` does NOT match;

Fixes #1136.

### DIFF
--- a/kytos/core/api_server.py
+++ b/kytos/core/api_server.py
@@ -17,6 +17,9 @@ from flask_cors import CORS
 from flask_socketio import SocketIO, join_room, leave_room
 from werkzeug.exceptions import HTTPException
 
+from kytos.core.auth import authenticated
+from kytos.core.config import KytosConfig
+
 
 class APIServer:
     """Api server used to provide Kytos Controller routes."""
@@ -296,6 +299,31 @@ class APIServer:
             return function
         return store_route_params
 
+    @staticmethod
+    def get_authenticate_options():
+        """Return configuration options related to authentication."""
+        options = KytosConfig().options['daemon']
+        return options.authenticate_urls
+
+    def authenticate_endpoints(self, napp):
+        """Add authentication to defined REST endpoints.
+
+        If any URL marked for authentication uses a function,
+        that function will require authentication.
+        """
+        authenticate_urls = self.get_authenticate_options()
+        for function in self._get_decorated_functions(napp):
+            inner = getattr(function, '__func__', function)
+            inner.authenticated = False
+            for rule, _ in function.route_params:
+                if inner.authenticated:
+                    break
+                absolute_rule = self.get_absolute_rule(rule, napp)
+                for url in authenticate_urls:
+                    if url in absolute_rule:
+                        inner.authenticated = True
+                        break
+
     def register_napp_endpoints(self, napp):
         """Add all NApp REST endpoints with @rest decorator.
 
@@ -315,8 +343,10 @@ class APIServer:
         for function in self._get_decorated_functions(napp):
             for rule, options in function.route_params:
                 absolute_rule = self.get_absolute_rule(rule, napp)
-                self._start_endpoint(napp_blueprint, absolute_rule, function,
-                                     **options)
+                if getattr(function, 'authenticated', False):
+                    function = authenticated(function)
+                self._start_endpoint(napp_blueprint, absolute_rule,
+                                     function, **options)
 
         # Register this Flask Blueprint in the Flask App
         self.app.register_blueprint(napp_blueprint)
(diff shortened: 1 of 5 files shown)


kytos/core/api_server.py
```diff
@@ -290,6 +293,34 @@ def store_route_params(function):
...
+    def get_authenticate_options():
+        """Return configuration options related to authentication."""
+        options = KytosConfig().options['daemon']
+        return options.authenticate_urls
+
+    def authenticate_endpoints(self, napp):
+        """Add authentication to defined REST endpoints.
+
+           If any URL marked for authentication uses a function,
+           that function will require authentication.
```
hdiogenes:
```suggestion
        If any URL marked for authentication uses a function,
        that function will require authentication.
```

kytos/core/config.py
```diff
@@ -148,6 +147,10 @@ def parse_args(self):
         self.parser.set_defaults(**defaults)
 
         self.options['daemon'] = self._parse_options(argv)
+        authenticate_urls = self.options['daemon'].authenticate_urls
+        self.options['daemon'].authenticate_urls = json.loads(
+            authenticate_urls
+            )
```
hdiogenes:
```suggestion
            authenticate_urls)
```

kytos/templates/kytos.conf.template
```diff
@@ -69,3 +69,10 @@ vlan_pool = {}
 
 # The jwt_secret parameter is responsible for signing JSON Web Tokens.
 jwt_secret = {{ jwt_secret }}
+
+# Define urls to be authenticated
```
hdiogenes:
```suggestion
# Define URLs that will require authentication
```

kytos/core/api_server.py
```diff
@@ -309,8 +340,11 @@ def register_napp_endpoints(self, napp):
         for function in self._get_decorated_functions(napp):
             for rule, options in function.route_params:
                 absolute_rule = self.get_absolute_rule(rule, napp)
-                self._start_endpoint(napp_blueprint, absolute_rule, function,
-                                     **options)
+                if hasattr(function, 'authenticated') and \
+                   function.authenticated:
```
hdiogenes:
```suggestion
                if getattr(function, 'authenticated', False):
```

kytos/core/config.py
```diff
@@ -125,8 +125,7 @@ def parse_args(self):
                     'port': 6653,
                     'foreground': False,
                     'protocol_name': '',
-                    'enable_entities_by_default': False,
-                    'napps_pre_installed': [],
+                    'authenticate_urls': '[]',
```
hdiogenes: Why were those two default values removed?

hdiogenes (changes requested):

kytos/core/config.py
ajoaoff: I think they shouldn't, but I don't remember why I did that. I will revert them.

ajoaoff:
I found a bug. Do not accept it yet.

hdiogenes (approved):

kytos/templates/kytos.conf.template
```diff
@@ -69,3 +69,10 @@ vlan_pool = {}
 
 # The jwt_secret parameter is responsible for signing JSON Web Tokens.
 jwt_secret = {{ jwt_secret }}
+
+# Define URLs that will require authentication
+#
+# This must a be a list of part of URLs. For example, if "kytos/mef_eline"
```
hdiogenes:
```suggestion
# This must be a list of part of URLs. For example, if "kytos/mef_eline"
```

kytos/core/api_server.py
```diff
@@ -290,6 +293,34 @@ def store_route_params(function):
...
+
+        If any URL marked for authentication uses a function,
+        that function will require authentication.
+        """
+        authenticate_urls = self.get_authenticate_options()
+        for function in self._get_decorated_functions(napp):
+            if hasattr(function, '__func__'):
+                inner = function.__func__
+            else:
+                inner = function
```
hdiogenes:
```suggestion
            inner = getattr(function, '__func__', function)
```

PR #1142, opened 2020-08-27
Create config field in Interface

### LINKED ISSUE #79: Kytos shouldn't send packet_out for Noviflow's BFD logical ports
In Noviflow switches, the logicalports type BFD does not accept OUTPUT action nor MATCH (but a group can be set up to watch a BFD logicalport).

How to detect this type of port? (needs info)

gleybersonandrade:
This PR creates a new 'config' field on Interface class. This field can assume one of the following values from Port or PhyPort classes:

OFPPC_PORT_DOWN: Port is administratively down;
OFPPC_NO_RECV: Drop all packets received by port;
OFPPC_NO_FWD: Drop packets forwarded to port;
OFPPC_NO_PACKET_IN: Do not send packet-in msgs for port.

Fix kytos/of_core#79

### DIFF
--- a/kytos/core/interface.py
+++ b/kytos/core/interface.py
@@ -79,10 +79,10 @@ class Interface(GenericEntity):  # pylint: disable=too-many-instance-attributes
                 to what is informed by the switch. Return ``None`` if not set
                 and switch does not inform the speed.
             config(|port_config|): Port config used to indicate interface
-                behavior (administratively down, ignore reveived packets, drop
-                forwarded packets, do not send packet-in messages). In general,
-                the port config bits are set by the controller and not changed
-                by the switch.
+                behavior. In general, the port config bits are set by the
+                controller and are not changed by the switch. Options
+                are: administratively down, ignore received packets, drop
+                forwarded packets, and/or do not send packet-in messages.
         """
         self.name = name
         self.port_number = int(port_number)
--- a/kytos/core/interface.py
+++ b/kytos/core/interface.py
@@ -63,7 +63,7 @@ class Interface(GenericEntity):  # pylint: disable=too-many-instance-attributes
 
     # pylint: disable=too-many-arguments
     def __init__(self, name, port_number, switch, address=None, state=None,
-                 features=None, speed=None):
+                 features=None, speed=None, config=None):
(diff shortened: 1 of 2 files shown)


gleybersonandrade:
More info about these flags can be seen on page 36 (Port Structures section) of Openflow 1.3 spec:

https://www.opennetworking.org/wp-content/uploads/2014/10/openflow-spec-v1.3.0.pdf

ajoaoff:
You stated that the field will assume one of the four values. In fact, each of the values is a bit, and the field can possibly have all of them set. For example, if the field has the value, in bits, `01000101`, then `OFPPC_NO_FWD` is `False` and the others are `True`.

kytos/core/interface.py
```diff
@@ -78,13 +78,19 @@ def __init__(self, name, port_number, switch, address=None, state=None,
             speed (int, float): Interface speed in bytes per second. Defaults
                 to what is informed by the switch. Return ``None`` if not set
                 and switch does not inform the speed.
+            config(|port_config|): Port config used to indicate interface
+                behavior (administratively down, ignore reveived packets, drop
+                forwarded packets, do not send packet-in messages). In general,
+                the port config bits are set by the controller and not changed
+                by the switch.
```
hdiogenes:
```suggestion
            config(|port_config|): Port config used to indicate interface
                behavior. In general, the port config bits are set by the
                controller and are not changed by the switch. Options
                are: administratively down, ignore received packets, drop
                forwarded packets, and/or do not send packet-in messages.
```

hdiogenes (changes requested):

ajoaoff (approved):

italovalcy:
Hi guys. THe PR seems to be very interesting for the switch management, in particular for those switches with special ports which does not support forwarding, packet_in, etc. I suppose this config should be changed only through OpenFlow PORT DESC messages, but not via Kytos REST API, right ? I'm asking this because I didn't see any modification on the REST API for this PR. Maybe it should be interesting to allow the administrator to manually change this config setting using the API (in case the switch reports wrong cur/max speed equal 0 - as in https://github.com/kytos/of_core/pull/94)

PR #1147, opened 2020-09-01
Update blueprints to include headers 

josemauro:


docs/blueprints/EP001.rst
```diff
@@ -1,3 +1,14 @@
+:EP: BP001
```
hdiogenes: Using both EP and BP for the ID is confusing. Also, some files have three digits, others two. I suggest leaving just the number, without padding:
```suggestion
:EP: 1
```

docs/blueprints/EP002.rst
```diff
@@ -1,13 +1,6 @@
...
-
-
-Status
-######
-
-Accepted
-
+:EP: BP002
+:Title: OpenFlow Handshake
+:Status: accepted
```
hdiogenes: As this is already an effort to standardize these headers, I think it would be nice to stick to one standard. :) E.g.: capitalized for Status.
```suggestion
:Status: Accepted 
```

hdiogenes (changes requested):
I've committed some changes to EP001 and added requests to standardize the headers. Otherwise it looks good to go.

docs/blueprints/EP001.rst
josemauro: Ok, I updated the others BP too.

docs/blueprints/EP002.rst
josemauro: Ok, I updated the others types of status too.

PR #1168, opened 2020-09-23
New kytosd option to run in foreground without terminal

### LINKED ISSUE #1161: New parameter to run Kytos in foreground mode without terminal
New parameter to the kytosd command line to run Kytos in foreground mode without a terminal.
This functionality is to address a problem running Kytos in Docker / Kubernetes in foreground mode causing the container to be immediately terminated.

rmotitsuki:
This PR add na new option to run kytosd in foreground  without terminal.
The main use is to run kytos in foreground mode in daemon containers like docker.

Fix #1161

### DIFF
--- a/kytos/core/config.py
+++ b/kytos/core/config.py
@@ -56,7 +56,7 @@ class KytosConfig():
                             action='store_true',
                             help="Run in foreground (ctrl+c to stop)")
 
-        parser.add_argument('-ft', '--no_terminal',
+        parser.add_argument('-F', '--no-terminal',
                             action='store_true',
                             help="Run in foreground without terminal (ctrl+c to stop).")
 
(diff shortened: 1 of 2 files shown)


kytos/core/config.py
```diff
@@ -56,6 +56,10 @@ def __init__(self):
                             action='store_true',
                             help="Run in foreground (ctrl+c to stop)")
 
+        parser.add_argument('-ft', '--no_terminal',
```
beraldoleal: `-ft` and `--no_terminal` are not an UNIX Posix standard. If we  *really* need this, I suggest something like `-F` and `--no-terminal`.

Not sure if I understood how this would run in the foreground without a terminal. Also if it is without a terminal how CTRL+C will stop?

Why not using nohup here?

beraldoleal (changes requested):
Hi @rmotitsuki thanks for this, but I left one comment for you

kytos/core/config.py
hdiogenes: Suggestions applied.

hdiogenes:
@rmotitsuki added suggestions, but the code is not working as expected: kytos just finishes.

Issue #1133, opened 2020-08-18
Can't Run Kytos in Docker/Kubernetes


Ktmi:
# Description

Running Kytos in Docker/Kubernetes does not work as expected.

Running Kytos in Daemon mode results in the container closing, due to the lack of an active foreground process. This could be circumvented with a wait loop, however if Kytos where to crash, the container would not automatically close.

Running Kytos in Foreground mode with `kytosd -f` results in the container closing due to the interactive terminating after being opened. The interactive terminal sees that it will not receive any input from standard input, and therefore terminates immediately.

# Recreating the Problem

To recreate this problem in Kubernetes, apply the following to a Cluster

```yml
apiVersion: v1
kind: Pod
metadata:
  name: kytos-test
spec:
  containers:
  - name: kytos
    image: kytos/nightly
    command: ["kytosd", "-f"]
    ports:
    - containerPort: 8181
    - containerPort: 6653
```

After applying the file above to a cluster, a Pod will be created within the cluster called `kytos-test`. It will be running a single container based on the `kytos/nightly` docker image, and will begin execution by running `kytosd -f`.

Checking the state of the pod with `watch kubectl get pods` shows that the pod enters the `Running` state, then enters the `Completed` state, then repeats until it reaches the `CrashLoopBackOff` state.

Running `kubectl describe pod/kytos-test` reveals that the exit code upon termination was 0, indicating no errors from Kytos.

Checking the logs of the terminated pod with `kubectl logs pod/kytos-test` shows that the program begins termination after starting the interactive terminal, producing output like the following:

```
2020-08-18 16:00:11,226 - INFO [kytos.core.logs] (MainThread) Logging config file "//etc/kytos/logging.ini" loaded successfully.
2020-08-18 16:00:11,227 - INFO [kytos.core.controller] (MainThread) //var/run/kytos
2020-08-18 16:00:11,228 - INFO [kytos.core.controller] (MainThread) Starting Kytos - Kytos Controller
2020-08-18 16:00:11,228 - INFO [kytos.core.controller] (MainThread) Starting TCP server: <kytos.core.atcp_server.KytosServer object at 0x7fd0b508cf98>
2020-08-18 16:00:11,228 - INFO [kytos.core.atcp_server] (MainThread) Kytos listening at 0.0.0.0:6653
[...]

PR #1170, opened 2020-10-05
Disable running interactive terminal by default with foreground mode

Ktmi:
# Description

This pull requests removes running the interactive terminal by default when executing in foreground mode,
and instead makes the feature accessible through the `-i` or `--interactive` command line option.

This pull resolves both #1168 and #1133

### DIFF
--- a/kytos/core/kytosd.py
+++ b/kytos/core/kytosd.py
@@ -145,7 +145,7 @@ def async_main(config):
 
     loop.call_soon(controller.start)
 
-    if controller.options.foreground:
+    if controller.options.interactive:
         executor = ThreadPoolExecutor(max_workers=1)
         loop.create_task(start_shell_async())
 
--- a/kytos/core/config.py
+++ b/kytos/core/config.py
@@ -56,6 +56,10 @@ class KytosConfig():
                             action='store_true',
                             help="Run in foreground (ctrl+c to stop)")
 
+        parser.add_argument('-i', '--interactive',
+                            action='store_true',
+                            help="Run with interactive terminal")
+
         parser.add_argument('-l', '--listen',
                             action='store',
                             help="IP/Interface to be listened")


kytos/core/kytosd.py
```diff
@@ -145,7 +145,7 @@ async def start_shell_async():
 
     loop.call_soon(controller.start)
 
-    if controller.options.foreground:
+    if controller.options.interactive:
```
beraldoleal, line 148: We don't need to change our default behavior here.

Using this patch, We are going to need an extra argument (`-i`) to run in interactive mode. This is changing how kytosd run and how a lot of scripts and users are expecting to run. Our documentation, tutorials everything is assuming this.

Also, I'm still not sure if we really need this extra parameter. There is no such option in `podman` to handle this ? Because, right now kytos is starting exactly as a POSIX daemon. Like, apache, mysql, and many other daemons.

Unless I'm missing something, I don't see a reason to change the default beahviour in kytos.

Let's try to reduce the problem to a podman or docker. (let's remove the upper layers that we have with kubernetes). Can you reproduce the problem only with docker/podman? There is any option inside podman that will proper handle this?

beraldoleal (changes requested):

Ktmi:
@beraldoleal 
Could you provide an example of a script, where this change would break the script.

hdiogenes:
@Ktmi, what @beraldoleal tried to explain is that because of a specific issue (trying to run Kytos inside Kubernetes) you're changing the general and expected behavior of `kytosd -f`, which means that it may not break any script, but it surely will break with the current users' expectations when running `kytosd -f`. That's why we won't accept this change, and why he suggested that we try to find a more specific solution to that problem.

hdiogenes:
Rejected because of the reasons explained above. #1168 can be accepted instead.

PR #1174, opened 2020-10-09
New note about the creation of UI folders

Niehaus:
The documentation seemed to be not so clear about not needed to create of all UI folders to work, the NApp just need the ones that are going to being used.

### DIFF
--- a/docs/developer/web-ui.rst
+++ b/docs/developer/web-ui.rst
@@ -201,6 +201,8 @@ have the following structure.
 When the administrator enables your NApp, Kytos will automatically scan the
 ``ui/*`` structure to properly export to GUI.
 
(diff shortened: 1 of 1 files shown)


docs/developer/web-ui.rst
```diff
@@ -201,6 +201,8 @@ have the following structure.
 When the administrator enables your NApp, Kytos will automatically scan the
 ``ui/*`` structure to properly export to GUI.
 
+.. note:: Kytos will parse the files structure and try to find your NApp components, so you need to create a folder only for the specific UI sections that are going to be used. So k-action-menu, k-info-panel and k-toolbar are the three UI areas available for using, feel free to create as you need it or not.
```
hdiogenes: Please try to keep the text at 80 columns when possible.

PR #1176, opened 2020-10-13
Create Blueprint for Constrained Shortest Path Search

MarvinTorres:
This blueprint is to add support for filtering by metadata using Kytos Pathfinder's `v2/` endpoint. In particular,
`shortest_path`, the method decorated with `@rest('v2/', methods=['POST']`, should get the user's edge attribute
requirements, remove best paths with edges that fail to meet those requirements, then return the remaining paths.
It covers details of the test that determines whether a path is preserved, flexibility, the dependencies that must be added, and 
the verification requirements.

It is a work in progress and is subject to change over time.

### DIFF
--- a/docs/blueprints/EP023.rst
+++ b/docs/blueprints/EP023.rst
@@ -1,0 +1,332 @@
+:EP: 23
+:Title: Kytos Pathfinder Filter Paths by Metadata
+:Authors:
+    - David Ramirez <drami102@fiu.edu>
+    - Marvin Torres <mtorr068@fiu.edu>
+:Created: 2020-10-05
+:Kytos-Version:
+:Status: Draft
+
+***************************************************
+EP023 - Kytos Pathfinder Filter Paths by Metadata
+***************************************************
+
+Abstract
+========
+
+This blueprint is to add support for filtering by metadata using Kytos Pathfinder's root `POST /` endpoint. In particular,
+`shortest_path`, the method decorated with `@rest('/vX', methods=['POST']`, should get the user's edge attribute
+requirements, remove best paths with edges that fail to meet those requirements, then return the remaining paths.
+It covers details of the test that determines whether a path is preserved, flexibility, the dependencies that must be added, and 
+the verification requirements.
+
+Motivation
+==========
+
+Currently, Kytos Pathfinder accepts the following parameters:
+ - Source
+ - Destination
+ - Desired links
+ - Undesired links
+ - Parameter
+
+*Desired links* and *undesired links* are noteworthy. They showcase Kytos Pathfinder's current filtering ability.
+Namely, they cause Kytos Pathfinder to remove edges that do not have all the desired links and none of the undesired ones. 
+
+In its current state, this list of parameters is insufficient for filtering out paths based on metadata information.
+Even *Parameter*, the only one that holds metadata information, can only hold one attribute name.
+
+In addition, Kytos Pathfinder's pathfinding ability is limited. While it can use weights, it assumes that the
+higher the weight, the worse its associated edge. While this is true for some attributes like delay and utilization,
+it is false for others like bandwidth. So finding the path with the highest bandwidth is currently impossible. Second,
+it can only find the best path in one area and not multiple - it can not find, for example, a path with ideal bandwidth
+and delay. Lastly, it finds the absolute best paths in the graph when the best might not be required. A user might
+look for a path that is good enough for their needs.
+
+Rationale
+=========
+
+These changes would add the following to the backend of Kytos Pathfinder: 
+ - **REST Endpoint**: The endpoint `/vX` (where X is an integer) that accepts the source, destination, and constraints, and outputs the constrained shortest path. 
+ - **Filter Functionality**: The ability to find the subgraph of the *graph* instance variable, namely one that contains only qualifying edges.
+
+Specification
+=============
+
+Details on testing a path and verifying these changes are discussed.
+
+Testing a Path
+--------------
+
+The path test assumes the following:
+
+Constraints were given by the user. For example:
+
+ - Bandwidth must be at least 34 units.
+ - Delay must be at most 45 units.
+ - Owner must be Joe.
+ - Reliability can be at least 3 units, or utilization can be at most 32 units, or both.
+ - Path must have links A and B.
+ - Path must not have link C.
+
+These are required because it will join attribute names in these constraints with the
+attribute names in an edge's metadata and see if the associated attribute values are
+high or low enough. 
+
+Also, it assumes that **all** edges in a graph have the following attributes:
+
+ - Bandwidth
+ - Delay
+ - Reliability
+ - Utilization
+ - Priority
+ - Ownership
+
+And that some attributes will have a maximum while others have a minimum:
+
+Minimum:
+ - Bandwidth
+ - Priority
+ - Reliability
+
+Maximum:
+  - Delay
+  - Utilization
+
+Equal:
+  - Ownership
+
+It will *pass* an edge if its metadata meets constraints, otherwise it will *fail* the edge. 
+Assume that a user wishes to check all edge attributes (checking all attributes is optional). First, it obtains 
+the constraint as a JSON object:
+
+  ::
+
+    "mandatory_filters": {
+      "bandwidth":A,
+      "delay":B,
+      "utilization":C,
+      "reliability":D,
+      "priority":E,
+      "ownership":F
+    }
+
+where **A** through **E** are floating point numbers or integers and **F** is a string.
+
+Then it retrieves an edge in the graph and performs the following test, which is a series of questions:
+
+ - bandwidth(edge) at least A?
+ - delay(edge) at most B?
+ - utilization(edge) at most C?
+ - reliability(edge) at least D?
+ - priority(edge) at least E?
+ - ownership(edge) equal to F?
+
+If YES to all questions, then it **passes** the edge.
+
+For example, assume that it receives the user constraint *Delay must be at most 45 units and bandwidth must be at least 30 units*.
+Then the incoming JSON object will be as follows:
+
+  .. code-block:: JSON
+
+    "mandatory_filters": {
+      "bandwidth":30,
+      "delay":45
+    }
+
+Then it will retrieve an edge and check if its bandwidth is at least 30 units and its delay
+is at most 45 units. If it does not have both qualities then it fails. Then it will repeat
+this for every subsequent edge it finds.
+
+A returned path will have all of the passing edges and none of the failing ones.
+
+To summarize, the input required is the user's constraint as a JSON object and the output is
+the set of paths that meets that constraint. 
+
+Flexibility in Theory
+---------------------
+
+Recall in the previous section that the test is a series of questions, and that a YES to all questions
+means a **passing result**.
+
+However, a user might be okay with some YESes. Support for flexibility can meet those needs.
+
+Flexible metrics are JSON objects, just like inflexible ones:
+
+  .. code-block:: JSON
+
+    "optional_filters": {
+      "bandwidth":A,
+      "delay":B,
+      "utilization":C,
+      "reliability":D,
+      "priority":E,
+      "ownership":F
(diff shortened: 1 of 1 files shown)


docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
+:EP: 22
+:Title: Kytos Pathfinder Filter Paths by Metadata
+:Authors:
+    - David Ramirez <drami102@fiu.edu>
+    - Marvin Torres <mtorr068@fiu.edu>
+:Created: 2020-10-05
+:Kytos-Version: 2020.2b
```
beraldoleal: `kytos-version` should be populated after this blueprint is in an approved state. You can keep it empty.

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+:Status: Draft
+
+Abstract
+========
+
+This blueprint is to add support for filtering by metadata using Kytos Pathfinder's `v2/` endpoint. In particular,
```
beraldoleal: v2 is a version endpoint, that will change often. What is the real endpoint?

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+it can only find the best path in one area and not multiple - it can not find, for example, a path with ideal bandwidth
+and delay. Lastly, it finds the absolute best paths in the graph when the best might not be required. A user might
+look for a path that is good enough for their needs.
+
+Rationale
+=========
+
+These changes would give users better control over which circuits to reserve. This is because they add the following abilities:
+ - **Constraint Menu**: Users can declare constraints when searching for paths. 
+ - **Flexibility Toggle**: Users can soften the constraints, so if a path cannot be found, find a path using less constraints than before.
```
beraldoleal: IIUC, to avoid a big changes, and a big PR, I suggest you focus here only on the backend part. We don't need to discuss menus/toggles on the web ui.

beraldoleal: IIUC, to avoid huge changes, and a big PR, my suggestion is to focus here only on the backend part (Rest endpoints and filter funcionalities). We don't need to discuss/mention menus/toggles that are part of the web ui.

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+
+*Desired links* and *undesired links* are noteworthy. They showcase Kytos Pathfinder's current filtering ability.
+Namely, they cause Kytos Pathfinder to remove edges that do not have all the desired links and none of the undesired ones. 
+
+In its current state, this list of parameters is insufficient for filtering out paths based on metadata information.
+Even *Parameter*, the only one that holds metadata information, can only hold one attribute name.
+
+In addition, Kytos Pathfinder's pathfinding ability is limited. While it can use weights, it assumes that the
+higher the weight, the worse its associated edge. While this is true for some attributes like delay and utilization,
+it is false for others like bandwidth. So finding the path with the highest bandwidth is currently impossible. Second,
```
beraldoleal: I'm not sure here, since it has some time that I did worked with this. But based on the code and on the REST API documentation is *not true*. Can you please double check? Like I said, I'm not sure here. BUT:

Based on the documentation, this endpoint return a *list* of paths. So multiple paths between two points will be returned, and if the order is ascending, users can reverse this list to get the right semanthics.

@ajoaoff , can you please double check here, too?

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+Equal:
+  - Ownership
+
+It will *pass* an edge if its metadata meets constraints, otherwise it will *fail* the edge. 
+Assume that a user wishes to check all edge attributes (checking all attributes is optional). First, it obtains 
+the constraint as a JSON object:
+
+  ::
+
+    "base_metrics": {
```
beraldoleal: My suggestion is to rename this dict name to something like "mandatory_filters".

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+Recall in the previous section that the test is a series of questions, and that a YES to all questions
+means a **passing result**.
+
+However, a user might be okay with some YESes. Support for flexibility can meet those needs.
+
+Flexible metrics are JSON objects, just like inflexible ones:
+
+  .. code-block:: JSON
+
+    "flexible_metrics": {
```
beraldoleal: same here, "optional_filters"

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+
+If YES to all questions, then it **passes** the edge.
+
+For example, assume that it receives the user constraint *Delay must be at most 45 units and bandwidth must be at least 30 units*.
+Then the incoming JSON object will be as follows:
+
+  .. code-block:: JSON
+
+    "base_metrics": {
+      "bandwidth":30,
```
beraldoleal: How the expression is going to be expressed here? Here I'm seeing only the numbers, I was expecting something like ">30".

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+All the passing links should be present in the first set, and all the failing links should be present
+in the second. 
+
+Possible Implementation
+=======================
+
+The following is two approaches to implement the endpoint change:
+
+  1. Edit the method. The downside to this is the work required to inform current Kytos Pathfinder users regarding the update. 
+  1. Write a new method with a new endpoint decorator. The downside to this is that other NApps have to be aware of its existence, increasing coupling.
```
beraldoleal: You can write new methods as you wish, no problem. But the REST API interface should be the same, for the requests, since this is *optional*. IMO we should not require one additional endpoint just to pass additional filters. Filters are data, and passed via POST.

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+
```
beraldoleal: I'm also missing here more details about the *how*. How do you plan to implement this? Today we have inside our main entities (Switch, Interface and Link) the `metadata` attribute. I' m assuming that those custom parameters are going to be as part of `link.metadata`. But not sure how this is going to work with networkx library, or if you are going to replace the networkx dependency that we have today.

beraldoleal (changes requested):
Hi @MarvinTorres , first of all thanks for this. I was not able to do a very detailed review here, but from the requirements point of view what you are proposing, it does make sense. Good job.

I left a few comments for you, but we should also wait for other comments. (@hdiogenes , @ajoaoff please make your comments here).

I will do a final review after another round of suggestions/changes.

docs/blueprints/EP022.rst
MarvinTorres: I can change the filter names to `min_bandwidth`, `max_delay`, `max_utilization`, `min_priority`, `min_reliability`, and `exact_owner`. That way the limits of a constraint are better specified than they are currently listed.

docs/blueprints/EP022.rst
MarvinTorres: Keeping the same API sounds good to me. It avoids the coupling issue and helps save time documenting the changes.

docs/blueprints/EP022.rst
MarvinTorres: I should specify that these changes can work with the networkx library because it supports link metadata. I will explain more in the updated blueprint.

docs/blueprints/EP022.rst
MarvinTorres: After the senior project, the endpoint was `v2/best-constrained-paths`. It is easy for me to change it to that. But I feel that dedicating a unique endpoint in general will require other NApps to distinguish endpoints and thus increase coupling overall.

As a result, I feel it may be best to keep it as `v2/` and its future iterations.

beraldoleal: I would suggest to not mention the v2, because this is a internal convention for endpoint versions. You are talking about the root endpoint "POST /". Yes, you can mention the version if you want, but we need to make clear here that his is about the root endpoint.

docs/blueprints/EP022.rst
beraldoleal: Based on your text, I think the flexibility is good in this case. So I still think that accepting an expression like ">", "<", "=", would be nice. Also, allow us to add new operators in the feature without changing the keys.

docs/blueprints/EP022.rst
beraldoleal: Ok, looking forward for a v2.

MarvinTorres:
I have posted the second rough draft of this blueprint.

Take a look and let me know what you think.

docs/blueprints/EP022.rst
ajoaoff: In fact, bandwidth cannot be used as a weight to find best paths, because the weight of a path is the sum of the weight of its edges, and that does not make sense with bandwidth. For bandwidth, it is the smallest edge in the path that will be the bandwidth of that path.

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,328 @@
+:EP: 22
+:Title: Kytos Pathfinder Filter Paths by Metadata
+:Authors:
+    - David Ramirez <drami102@fiu.edu>
+    - Marvin Torres <mtorr068@fiu.edu>
+:Created: 2020-10-05
+:Kytos-Version:
+:Status: Draft
+
```
Niehaus: To build the blueprints table and its navigation properly, we have a new standard in EP000 for the title as follows: 

```suggestion
***************************************************
EP022 - Kytos Pathfinder Filter Paths by Metadata
***************************************************

```

Niehaus (changes requested):

docs/blueprints/EP022.rst
MarvinTorres: Thank you for the heads up! I will commit your suggestion.

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,331 @@
+:EP: 22
+:Title: Kytos Pathfinder Filter Paths by Metadata
+:Authors:
+    - David Ramirez <drami102@fiu.edu>
+    - Marvin Torres <mtorr068@fiu.edu>
+:Created: 2020-10-05
+:Kytos-Version:
+:Status: Draft
```
hdiogenes: New line needed here for the document to build correctly.

```suggestion
:Status: Draft

```

docs/blueprints/EP022.rst
josemauro: Use "fewer" instead "less".

docs/blueprints/EP022.rst
```diff
@@ -0,0 +1,256 @@
...
+All the passing links should be present in the first set, and all the failing links should be present
+in the second. 
+
+Possible Implementation
+=======================
+
+The following is two approaches to implement the endpoint change:
+
+  1. Edit the method. The downside to this is the work required to inform current Kytos Pathfinder users regarding the update. 
+  1. Write a new method with a new endpoint decorator. The downside to this is that other NApps have to be aware of its existence, increasing coupling.
```
josemauro: Please check the enumeration, there are two itens "1".

josemauro (changes requested):

docs/blueprints/EP022.rst
MarvinTorres: Thank you for the heads up! Will fix.

docs/blueprints/EP022.rst
MarvinTorres: @josemauro This seems to be from an outdated version of this section.

docs/blueprints/EP022.rst
MarvinTorres: @josemauro I dug into this a bit and found that I rewrote these sections a while ago. The rewritten sections do not have the issues you pointed out.

hdiogenes:
Document is not building correctly: `/home/scrutinizer/build/docs/blueprints/EP023.rst:156:Could not lex literal_block as "JSON".`

hdiogenes:
Superseded by #1252

PR #1177, opened 2020-10-14
Use new pip dependency resolver

hdiogenes:


hdiogenes:
Closing as it currently doesn't work on Scrutinizer (old pip used by it doesn't support the new flag).

PR #1196, opened 2020-10-26
Update k-action-menu event documentation

Niehaus:
Updated the documentation about k-action-menu event and added a example of usage with a `statusMessage` in the kytos UI


docs/developer/events.rst
```diff
@@ -1,3 +1,7 @@
+Kytos have some events available, and to call them you should use the sintaxe
+`this.$kytos.$emit()` passing as parameters the specific events names and its
+objects of acceptance as documented bellow:
```
hdiogenes:
```suggestion
Kytos has some events available, and to call them you should use the syntax
`this.$kytos.$emit()` passing as parameters the specific events names and its
objects of acceptance as documented below:
```

Issue #1235, opened 2021-04-03
Intermittent RuntimeError when killing kytosd


rmotitsuki:
Killing kytosd process using command `pkill kytosd` throws a RuntimeError in log.
It is intermittent and happens with foreground and background kytos.
It is not critical to Kytos normal functioning, but it causes errors in end-to-end test that monitors kytos log files.

```
concurrent.futures:ERROR _base:326:  exception calling callback for <Future at 0x7f63c09897b8 state=finished returned NoneType>
Traceback (most recent call last):
File "/usr/lib/python3.6/concurrent/futures/_base.py", line 324, in _invoke_callbacks                               
  callback(self)
File "/usr/lib/python3.6/asyncio/futures.py", line 417, in _call_set_state  
  dest_loop.call_soon_threadsafe(_set_state, destination, source)                                        
File "/usr/lib/python3.6/asyncio/base_events.py", line 637, in call_soon_threadsafe           
  self._check_closed()                                                                                            
File "/usr/lib/python3.6/asyncio/base_events.py", line 377, in _check_closed
  raise RuntimeError('Event loop is closed')                                                               
```

hdiogenes:
Fixed in #1238.

PR #1238, opened 2021-04-13
Fix Runtime errors when closing Kytos 

rmotitsuki:
Fix issue #1235 

### :bookmark_tabs: Description of the Change

Fixed kytosd shutdown to cancel tasks before closing the event loop that executes the tasks.
Closing the event loop while tasks are being completed causes Future Runtime errors.

### :computer: Verification Process

Verified using the command `pkill kytosd` with kytos in daemon and foreground modes.

### DIFF
--- a/tests/unit/test_core/test_kytosd.py
+++ b/tests/unit/test_core/test_kytosd.py
@@ -2,7 +2,7 @@
 from unittest import TestCase
 from unittest.mock import MagicMock, patch
 
-from kytos.core.kytosd import _create_pid_dir, async_main, main, start_shell
+from kytos.core.kytosd import _create_pid_dir, async_main, main, create_shell
 
 
 class TestKytosd(TestCase):
@@ -26,14 +26,15 @@ class TestKytosd(TestCase):
         (mock_mkdirs, mock_chmod) = args
         _create_pid_dir()
 
-        mock_mkdirs.assert_called_with('/var/run/kytos', exist_ok=True)
+        mock_mkdirs.assert_called_with('/var/run/kytos', exist_ok=Tre)
         mock_chmod.assert_called_with('/var/run/kytos', 0o1777)
 
     @staticmethod
     @patch('kytos.core.kytosd.InteractiveShellEmbed')
     def test_start_shell(mock_interactive_shell):
         """Test stop_api_server method."""
-        start_shell(MagicMock())
+        ipshell = create_shell(MagicMock())
+        ipshell()
 
         mock_interactive_shell.assert_called()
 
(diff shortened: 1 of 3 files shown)


hdiogenes:
@rmotitsuki the asyncio API has changed in Python 3.9. I tried to make the code compatible with the new versions (as per [this Jupyter example](https://github.com/jupyterhub/jupyterhub/pull/3298/files)), but I'm receiving this error now:

```
2021-04-30 12:45:18,950 - INFO [kytos.core.controller] (MainThread) Shutting down NApp kytos/of_core...
2021-04-30 12:45:18,961 - INFO [kytos.core.api_server] (MainThread) The Rest endpoints from /api/kytos/of_core/ were disabled.
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
~/src/kytos/kytos/bin/kytosd in <module>
      5 __file__ = '/Users/humberto/src/kytos/kytos/bin/kytosd'
      6 with open(__file__) as f:
----> 7     exec(compile(f.read(), __file__, 'exec'))

~/src/kytos/kytos/bin/kytosd in <module>
      4 from kytos.core import kytosd
      5
----> 6 kytosd.main()

~/src/kytos/kytos/kytos/core/kytosd.py in main()
    105
    106     if config.foreground or not config.daemon:
--> 107         async_main(config)
    108     else:
    109         with daemon.DaemonContext():

~/src/kytos/kytos/kytos/core/kytosd.py in async_main(config)
    168         controller.log.info("Shutting down Kytos...")
    169     finally:
--> 170         pending = [t for t in asyncio_all_tasks() if
    171                    t is not asyncio_current_task()]
    172         for task in pending:

/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/tasks.py in all_tasks(loop)
     42     """Return a set of all tasks for the loop."""
     43     if loop is None:
---> 44         loop = events.get_running_loop()
     45     # Looping over a WeakSet (_all_tasks) isn't safe as it can be updated from another
     46     # thread while we do so. Therefore we cast it to list prior to filtering. The list

RuntimeError: no running event loop
```

tests/unit/test_core/test_kytosd.py
```diff
@@ -26,14 +26,15 @@ def test_create_pid_dir__system(*args):
         (mock_mkdirs, mock_chmod) = args
         _create_pid_dir()
 
-        mock_mkdirs.assert_called_with('/var/run/kytos', exist_ok=True)
+        mock_mkdirs.assert_called_with('/var/run/kytos', exist_ok=Tre)
```
hdiogenes:
```suggestion
        mock_mkdirs.assert_called_with('/var/run/kytos', exist_ok=True)
```

